Issue #3, opened 2021-02-11
pyCPA port: core.combine_rows


mikapfl:
Port the functionality of pyCPA's [core.combine_rows](https://github.com/JGuetschow/pyCPA/blob/master/src/pyCPA/core/core.py#L372).

mikapfl:
`combine_rows` functionality:
1. Restrict the data by indexing on specified dimensions (defined
   in `other_cols`).
2. Along a specified dimension either of (defined in `mapping`):
   - `x = sum(a, b, c, d)`
   - `x = a - sum(b, c, d)`
   - `x = a`
   - `x = -a`
3. Multiple dimensions can be specified in the `mapping`, the actions will
   be applied simultaneously. With the given operators, this is
   mathematically equivalent to applying them serially.
4. Metadata dimensions are dropped (defined in `cols_to_remove`).

mikapfl:
If `x` is a new value for the dimension, in xarray we need to reindex to insert it. Maybe we want a function where the caller doesn't have to care if a particular value along a dimension already exists or not, I think then all the functionality can be reached by a few chained standard xarray calls.

JGuetschow:
comments on `combine rows` functionality:
1. yes
2. all combinations of + and - are possible (e.g. `x = - sum(a,b,c)`).
3. yes
4. yes, but maybe that's not necessary because the metadata is stored in a different way in primap2

And regarding the next comment: I think it would be good if one does not have to know if a value already exists. 

For the data we need user definable rules on what to do in that case (keep old, overwrite, ...)

PR #12, opened 2021-02-16
Setter methods for easier manipulation / extension

mikapfl:
# Pull request

- [x] Tests added

## Description

Add a method to set values which works both when the new values are missing beforehand or when they exists. In effect, bundles `xarray`'s `reindex`, `fillna`, and `combine_first` functionality together.

Is part of the effor to replace pyCPAs `combine_rows` functionality (#3 )


mikapfl:
To see the built documentation without building it locally, press "show all checks" beneath "All checks have passed" in the infobubble and then click on "Details" on the `docs/readthedocs.org:primap2` check.

docs/usage.ipynb
```diff
@@ -228,6 +229,232 @@
     }
    }
   },
+  {
```
JGuetschow, line 232: notebooks are hard to review

mikapfl: true, probably better to review the compiled section in the docs. /-:

primap2/_data_format.py
```diff
@@ -251,7 +261,7 @@ def ensure_entity_and_units_valid(key: Hashable, da: xr.DataArray):
                 f"compatible with an emission rate."
             )
     except pint.UndefinedUnitError:
-        if entity not in ("population",) and "(" not in entity:
+        if entity not in ("population",) and "(" not in key:
```
JGuetschow, line 264: Is this a workaround because population is not available in scmdata-units yet?

primap2/_downscale.py
```diff
@@ -11,12 +11,12 @@ class DataArrayDownscalingAccessor(BaseDataArrayAccessor):
     def downscale_timeseries(
         self,
         *,
-        dim: str,
-        basket: str,
-        basket_contents: Sequence[str],
+        dim: Hashable,
```
JGuetschow, line 14: What does changing the type from str to Hashable mean in practice?

primap2/_data_format.py
mikapfl: That's a philosphical question. According to metrologists, "population" does not have a unit, so it is fine to be unit-less here. If scmdata-units should gain a unit suitable for the entity "population", we would of course use it.

primap2/_downscale.py
mikapfl: The main reason to do it is that all the xarray-Functions take `Hashable`, so it makes mypy happy. In theory, it is possible to use any hashable type (for example, an int, or a tuple of strings) as a dimension name for xarray. In practice, we should never ever use anything else than a str, I think. I tried writing a test function with an int dimension name, and something inside xarray breaks before primap2 can even start complaining.

I liked `str` more because it is more explicit and we don't support other Hashables, but even though `str`s are a subset of Hashables, somehow static type inspection is only happy with `Hashable`. :confused:

primap2/_setters.py
```diff
@@ -0,0 +1,516 @@
...
+        dim: typing.Hashable,
+        key: typing.Any,
+        value: xr.Dataset,
+        existing: str,
+    ) -> xr.DataArray:
+        if dim not in da.dims:
+            return da
+        return da.pr.set(dim=dim, key=key, value=value[da.name], existing=existing)
+
+    def set(
```
JGuetschow, line 313: How can set be limited to some or just one variable(s)?

primap2/tests/test_setters.py
```diff
@@ -0,0 +1,383 @@
+"""Tests for _setters.py"""
```
JGuetschow, line 1: I have to learn pytest before I can say much about the tests

JGuetschow (approved):
I've looked at the code and all looks fine, however, I could not check everything in detail. I can't say much about the tests as I have to learn pytest first

primap2/_units.py
```diff
@@ -259,29 +263,40 @@ def quantify(self, units=None, **unit_kwargs) -> xr.Dataset:
...
+            a        (x) int64 0 3 2
+            b        (x) int64 5 -2 1
 
         >>> ds.pr.quantify()
         <xarray.Dataset>
         Dimensions:  (x: 3)
         Coordinates:
           * x        (x) int64 0 1 2
-            u        (x) int64 <Quantity([-1  0  1], 'second')>
+            u        (x) int64 [s] -1 0 1
```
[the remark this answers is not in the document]
mikapfl: Yes, I copied those examples from pint (which is the reason why we have the licenses/ folder already) and they were old, from pre-0.1.0 days. I updated them here and upstream.

primap2/tests/conftest.py
```diff
@@ -26,29 +28,7 @@ def emit(self, record):
...
-            for ent in ("CO2", "SF6", "CH4")
-        },
-        attrs={"area": "area (ISO3)"},
-    ).pr.quantify()
-
-    with ureg.context("SARGWP100"):
-        minimal["SF6 (SARGWP100)"] = minimal["SF6"].pint.to("CO2 Gg / year")
-    minimal["SF6 (SARGWP100)"].attrs["gwp_context"] = "SARGWP100"
-    return minimal
+    return examples.minimal_ds()
```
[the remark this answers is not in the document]
mikapfl: Yeah, I wanted to use the minimal dataset somewhere else (in the usage examples or so). Makes sense to also move the opulent one, will do.

primap2/_setters.py
mikapfl: Given a Dataset with multiple variables, limiting set to e.g. "CO2" would be done like this:
```python
ds["CO2"] = ds["CO2"].pr.set(…)
```
i.e. use the variable's set method instead of the Datasets set method.

primap2/tests/test_setters.py
mikapfl: the most interesting concepts are fixtures and pytest.mark.parametrize, the rest of pytest is pretty straight-forward, I think.

primap2/tests/utils.py
```diff
@@ -13,5 +13,25 @@ def allclose(a: xr.DataArray, b: xr.DataArray, *args, **kwargs):
 
 def assert_equal(a: xr.DataArray, b: xr.DataArray, *args, **kwargs):
     assert allclose(a, b, *args, **kwargs)
-    assert a.attrs == b.attrs
-    assert a.name == b.name
+    assert a.attrs == b.attrs, (a.attrs, b.attrs)
+    assert a.name == b.name, (a.name, b.name)
+
+
+def assert_align(a: xr.DataArray, b: xr.DataArray) -> (xr.DataArray, xr.DataArray):
```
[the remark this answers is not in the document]
mikapfl: agreed, I have added some documentation to the helper test functions meant for re-use

primap2/_data_format.py
JGuetschow: It would make sense to have the million etc prefixes though, I think.
In primap 1 units is "pers"

primap2/_downscale.py
JGuetschow: OK, got it

JGuetschow:
fro my side this can be merged now

mikapfl:
The RSE people wanted to have a look at it as well regarding structuring of the repo and the API, and given they have also built one or two data classes, I think that can be useful. Will happen likely on monday, so I'll merge on monday.

JGuetschow:
It's great if they can help.

mikapfl:
The RSE people noted that it would also be great to have the option to raise an error if a category does not exist yet. This is useful when you know you want to modify an existing category, and want to guard against e.g. typos. From an API side, this would be largely equivalent to:
```python
da.pr.loc[{dim: key}] = value
```
With the differences that
1. it uses the same API as setting a potentially new `key`
2. it doesn't modify the values in place
3. No `[..., np.newaxis]`-reordering is necessary depending on the position of `dim` in `da.dims`.

I think that this would indeed be very valuable. My initial idea would be to add another parameter like `existing`, probably called `new`, which by default would have a value of `new="extend"`, and an optional value of `new="error"`. Then, these two calls would be equivalent:
```python
da.pr.loc[{dim: key}] = value
da = da.pr.set(dim, key, value, new="error", existing="overwrite")
```
Of course, also `new="error", existing="fillna_empty"` (or just leaving `existing` out) could be quite useful. The only one which doesn't make much sense would be `new="error", existing="error"`, which would raise an error no matter what. :stuck_out_tongue_closed_eyes:

JGuetschow:
Would this check be on a dataset wide level or by time series?
So if a category exists for one country but not for the other would an error be raised on not? (I think it should not be raised)

mikapfl:
> So if a category exists for one country but not for the other would an error be raised on not? (I think it should not be raised)

Can never happen, in Datasets, the axes for all data variables are identical. So if a category exists for one country, it exists for all gases for all countries (it might contain NaN, but it exists, which this check would be about).

JGuetschow:
makes sense, I'm still thinking in terms of timeseries with metadata not multi-dimensional arrays.

mikapfl:
I'm not so sure now if ``new="extend"`` (add new keys to the dimension) or ``new="error"`` (raise a `KeyError` if keys are missing from the dimension) make sense as a default. On one hand, a default of `new="error"` is safer (guards against typos) and more explicit. On the other hand, for my mental model, it doesn't make much difference if a key exists and is filled by NaNs or if it doesn't exist, so for my mental model, `new="extend"` would make more sense as a default.

JGuetschow:
I think new="extend" is better for me as that's what I would mostly use

PR #31, opened 2021-03-24
Py cpa code integration

JGuetschow:
# Pull request

- [x] Tests added

## Description

Add read and write functions for the interchange format


mikapfl:
The test is failing at the moment because it is not yet specified what to do with datasets with different dimensions for different entities.

In the interchange format, all dimensions have to be specified, so e.g. the `category` has to be specified for the `population` in the opulent_ds. At the moment, the function simply broadcasts all entities to all dimensions, so the population is copied and the same value is given for all categories. That's stupid of course, but what would be better?

JGuetschow:
I didn't think about different dimensions for different variables when designing the interchange format, because I came from the wide csv file direction. It could be an option to write individual csv files for each set of variables with the same dimensions. Or just fill the missing dims with a default value that is discarded on reading.

mikapfl:
I'm now writing a definition for the interchange format which supports all features of primap2 (apart from additional coordinates, for now). However, in doing so it is becoming apparent for me, that we shouldn't advertise it as "the format that we publish in", instead we should say (IMO) something like "we publish in netcdf format, but for the benefit of users who need tabular data, we also publish in the interchange format". Honestly, netcdf was developed in a 30-year period (mainly by the geosciences community) and it is simply vastly superior to CSVs, and we should publish in this well-defined format for those users who can use it (because they are using R/python/julia/matlab/Fortran/C/any other language which has good netcdf libraries) and additionally in CSV format for those who have to use Excel or other strictly tabular tools.

JGuetschow:
I think publishing in netcdf format is almost pointless, because most of the users of the data already have problems using csv files which is why we publish some datasets also as xls files. We can of course try to motivate people who use programming languages and not excel to use netcdf by publishing in that format, but currently I doubt anyone would use it. But as I'm usually pro trying to change stuff, I'm also fine with saying that netcdf is the main format. After all it's no additional work to publish in netcdf (and one processing step less when using our data with out software).

JGuetschow:
I've just converted PRIMAP-crf to IF and netcdf. The csv has 7MB, the netcdf file 124MB. netcdf with compression is 6.2MB. Somewhow ntcdf doesn't seem to be the most efficient way to store the data (because of the sparseness of the data?).

mikapfl:
Yeah, sparseness. That's why compression is so efficient. I'd always use compression.

mikapfl:
Probably we could make `pr.to_netcdf()` use compression by default. zlib compression deals nicely with sparseness and is supported everywhere.

mikapfl:
compression is a bit awkward, because you can use different compression algorithms for the different entities (mainly to use differential compression for entities that are very similar for all data points). I guess we'll usually not be in the need to fine-tune the compression to this level usually, so we could a "just zlib" default.

mikapfl:
I started the documentation + implementation of a version of the interchange format with all bells and whistles (explicitly defined variables and dimensions per variable, explicitly defined time format, etc.) over in https://github.com/mikapfl/primap2/tree/interchange_format .

Unfortunately, I'll also have to change the data reading functions so that they generate all of the needed meta data, I guess I'll do it on monday.

Somehow just porting a single CSV data reading function escalated quickly, collectively we probably already spent something like 40 hours on this pull request.

PR #39, opened 2021-05-03
add functionality to convert long format dataframe (with non-PRIMAP) …

JGuetschow:
…columns and metadata to interchnage format

small changes and fixed to primap1 related mapping functions (allow more flexible unit input, map entities before units to ensure correct unit mapping)

# Pull request

- [x] Tests added

## Description

### DIFF
--- a/primap2/pm2io/_data_reading.py
+++ b/primap2/pm2io/_data_reading.py
@@ -591,12 +591,14 @@ def convert_wide_dataframe_if(
     else:
         time_columns = time_cols
 
-    filter_data(data_wide, filter_keep, filter_remove)
+    # make a copy of the data to not alter the input data
+    data_if = data_wide.copy()
+    filter_data(data_if, filter_keep, filter_remove)
 
-    add_dimensions_from_defaults(data_wide, coords_defaults)
+    add_dimensions_from_defaults(data_if, coords_defaults)
 
     naming_attrs = rename_columns(
-        data_wide, coords_cols, add_coords_cols, coords_defaults, coords_terminologies
+        data_if, coords_cols, add_coords_cols, coords_defaults, coords_terminologies
     )
     attrs.update(naming_attrs)
 
@@ -605,31 +607,30 @@ def convert_wide_dataframe_if(
     )
 
     if coords_value_mapping is not None:
-        map_metadata(data_wide, attrs=attrs, meta_mapping=coords_value_mapping)
+        map_metadata(data_if, attrs=attrs, meta_mapping=coords_value_mapping)
 
     if coords_value_filling is not None:
-        data_wide = fill_from_other_col(
-            data_wide, attrs=attrs, coords_value_filling=coords_value_filling
+        data_if = fill_from_other_col(
+            data_if, attrs=attrs, coords_value_filling=coords_value_filling
         )
 
-    coords = list(set(data_wide.columns.values) - set(time_columns))
-    # dims = list(set(coords) - set(add_coords_cols.keys()))
+    coords = list(set(data_if.columns.values) - set(time_columns))
 
-    harmonize_units(data_wide, dimensions=coords, attrs=attrs)
+    harmonize_units(data_if, dimensions=coords, attrs=attrs)
 
-    data_wide, coords = sort_columns_and_rows(data_wide, dimensions=coords)
+    data_if, coords = sort_columns_and_rows(data_if, dimensions=coords)
     dims = coords.copy()
     for add_coord in add_coords_cols.keys():
         dims.remove(add_coord)
 
-    data_wide.attrs = interchange_format_attrs_dict(
+    data_if.attrs = interchange_format_attrs_dict(
         xr_attrs=attrs,
         time_format=time_format,
         dimensions=dims,
         additional_coordinates=additional_coordinates,
     )
 
-    return data_wide
+    return data_if
 
 
 def read_wide_csv_file_if(
@@ -1246,7 +1247,7 @@ def rename_columns(
 def harmonize_units(
     data: pd.DataFrame,
     *,
-    unit_col: str = "unit",
+    unit_col: str = None,
     attrs: Optional[dict] = None,
     dimensions: Iterable[str],
 ) -> None:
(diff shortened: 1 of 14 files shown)


primap2/pm2io/_data_reading.py
```diff
@@ -41,6 +41,178 @@
...
+def convert_long_dataframe_if(
+    data_long: pd.DataFrame,
+    *,
+    coords_cols: Dict[str, str],
+    coords_defaults: Optional[Dict[str, Any]] = None,
+    coords_terminologies: Dict[str, str],
+    coords_value_mapping: Optional[Dict[str, Any]] = None,
+    coords_value_filling: Optional[Dict[str, Dict[str, Dict]]] = None,
+    filter_keep: Optional[Dict[str, Dict[str, Any]]] = None,
+    filter_remove: Optional[Dict[str, Dict[str, Any]]] = None,
```
mikapfl, line 53: filter_keep and filter_remove don't make much sense here, I think. If I have a dataframe already, I would filter it separately before calling `convert_long_dataframe_if`, I think.

primap2/pm2io/_data_reading.py
```diff
@@ -41,6 +41,178 @@
...
+        PRIMAP2 metadata.
+
+    coords_value_filling : dict, optional
+        A dict with primap2 dimension names as keys. These are the target columns where
+        values will be filled (or replaced). Vales are dicts with primap2 dimension names
+        as keys. These are the source columns. The values are dicts with source value -
+        target value mappings.
+        This can be used to e.g. fill missing category codes based on category names or
+        to replace category codes which do not meet the terminology using the category
+        names.
```
mikapfl, line 105: I think I don't understand what `coords_value_filling` can do what `coords_value_mapping` can't do. I think that should be spelled out in the documentation

primap2/pm2io/_data_reading.py
```diff
@@ -539,6 +711,64 @@ def filter_data(
...
+
+    attrs : dict
+        Dataset attributes
+
+    Returns
+    -------
+    pd.DataFrame
+    """
+    df_filled = df.copy()
```
mikapfl, line 748: The copying is not needed, I think. We don't copy in any of the mapping functions, either, so I think we can just operate in place here as well.

mikapfl (commented):
Looks good! Maybe we can also then re-write `read_long_csv_file_if` to be a very thin wrapper around `convert_long_dataframe_if` to have less duplicate code?

primap2/pm2io/_data_reading.py
JGuetschow: My intention was to put all code that is the same for all long format dataframes into the function to keep custom scripts as short as possible and the functionality as close to the read_XXX functions as possbile

mikapfl:
For seeing how it works out I added a commit to which makes `read_long_csv_file_if` a very thin wrapper around `convert_long_dataframe_if`. Seems to work nicely.

primap2/pm2io/_data_reading.py
JGuetschow: OK, will clarify that. 
The idea is that sometimes you have non-unique category codes but unique descriptions. This can only be fixed by using information from one column on another column. But coords_value_filling can actually replace coords_value_mapping. Just that the format is a bit more complicated.

JGuetschow: It's a functionality that was missing in PRIMAP1 and complicated data reading in some cases

primap2/pm2io/_data_reading.py
mikapfl: Yeah, keeping the API as close as possible is a good reason, true. Individually, it might make a lot of sense for performance reasons to filter earlier, but users are free to do that and ignore the `filter_*` parameters, so it makes total sense to keep them.

primap2/pm2io/_data_reading.py
JGuetschow: OK, will change that. It's probably a leftover from testing in a notebook

JGuetschow: done

primap2/pm2io/_data_reading.py
JGuetschow: added a bit of documentation where the function is used. We could integrate it also into other reading functions. And we have to decide if we replace `coords_value_mapping`.

mikapfl: Cool, now I understand the functionality. I like the possibility, because this is already what is needed for units sometimes (because you also need the entity) and so far there is a hack which makes it possible for built-in coord mappings like `PRIMAP1`, but not flexibly for coord mappings supplied by the user. Would have to think how we can preserve the functionality of `coords_value_mapping` without making everything too complicated.

JGuetschow: Currently information from different columns can not be combined, which is necessary for the PRIMAP1 to PRIMAP2 unit conversion.

mikapfl: Hmm... I'm thinking about a somewhat nice API which is not completely confusing...

Using introspection, function pointers could be used for a quite nice API, like this:

```python
def calculate_unit(entity, unit, category):
    """just a silly example for a complex multi-category mapping"""
    if entity == "KYOTOGHG":
        return "CO2"
    elif unit == "kt" and category == "average speed of international marine trade":
        return "knots"
    else:
        return unit

coords_value_mapping={"unit": calculate_unit})
```
Where the function can then extract the needed inputs from the parameter names (in the example, `unit`, `entity`, and `category`). Would be kind of straight forward and easy to understand without deeply nested dictionaries. Does that make sense? Or is it too verbose for the "I just want to fill one column from another column" use case?

JGuetschow:
Still missing:
* Tests for error messages for additional_coordinates reading
* Tests for fill_from_other_col
* Tests for the functions in _GHG_inventory_reading.py

primap2/pm2io/_data_reading.py
JGuetschow: the problem I see here is that you need harmonized column names to re-use the function. But if it's called in the right place (directly before conversion to IF) and after renaming this should be possible (when using the column name aliases as in the xarray format). I don't think it's too confusing and one could still use the old dict input, or provide a template for a function that does the simple replacing

JGuetschow:
can this be merged now, or are there open points to discuss first?

mikapfl:
For me it is fine. Maybe there was confusion, because the PR is still marked as "work in progress", so I thought you wanted to do something first, and maybe you thought I wanted to do something first. (%

Anyway, fine for me to merge

JGuetschow:
Just forgot to mark it as ready for review

PR #90, opened 2021-12-22
Integrate fixes and improvements to data reading

JGuetschow:
# Pull request

- [x] Tests added

## Description

This pull request mainly add a few bugfixes to data reading functionality. Additional changes are
* more flexibility in input format to `convert_ipcc_code_primap_to_primap2` (tests have been extended and adjusted to new error messages)
* data conversion function now work on a copy of the input DataFrame to avoid changes to the input DataFrame
*


primap2/pm2io/_data_reading.py
```diff
@@ -1289,26 +1340,97 @@ def harmonize_units(
...
+                    except TypeError:
+                        # print(data.loc[mask, data_cols])
+                        temp = []
+                        for col in data_cols:
+                            temp = temp + list(data[col].unique())
+                        temp = list(set(temp))
+
+                        # print(temp)
+                        strs = [x for x in temp if isinstance(x, str)]
+                        print(strs)
```
mikapfl, line 1427: This is effectively a `pass` on `TypeError`, and the function will soldier on. Maybe at least re-raise?

primap2/pm2io/_data_reading.py
```diff
@@ -1289,26 +1340,97 @@ def harmonize_units(
...
+                # we have a gas basket or something unknown, so no conversion to native
+                # unit
+                # print(f"Exception occured for entity {entity}")
+                unit_to = unit_fallback
+                entity_conv = False
+                native_unit = "None"
+
+            # print(f"unit_to is {unit_to}, unit_fallback is {unit_fallback}, and
+            # native_unit is {native_unit}")
+            # print(f"Using {unit_to} as target unit and entity_conv = {entity_conv}")
```
mikapfl, line 1397: I have moved this to a separate function to keep things tidy and easily tested.

primap2/pm2io/_GHG_inventory_reading.py
```diff
@@ -34,11 +38,16 @@ def nir_add_unit_information(
...
+        integer specifying the row to use for entity information.
+        If entity and unit information are given in the same row use only unit_row
     unit_info : dict
         A dict with the fields
         * regexp_entity: regular expression that extracts the entity from the cell value
         * regexp_unit: regular expression that extracts the unit from the cell value
-        * manual_repl: optional: dict defining unit and entity for given cell values
+        * manual_repl_unit: optional: dict defining unit for given cell values
+        * manual_repl_entity: optional: dict defining entity for given cell values
         * default_unit: unit to be used if no unit can be extracted an no unit is given
```
mikapfl, line 51: It would be much better to have all these as separate parameters. There is no reason to hide them in a dict (it is not passed around further) and using separate parameters, we can use type hints etc.

primap2/pm2io/_data_reading.py
```diff
@@ -139,6 +163,10 @@ def convert_long_dataframe_if(
         in the interchange format.
         Default: "%F", i.e. the ISO 8601 date format.
 
+    convert_nan : bool, optional
+        If set to true, string values in the data columns as listed in NA_VALUES will be
```
mikapfl, line 167: For the user, NA_VALUES is not readily available, so the documentation here is rather unclear. Either just use something like "common values denoting NaN" or export the NA_VALUES and add the full name (`primap2.pm2io.NA_VALUES` or so) or maybe even make it possible to supply an own list and use NA_VALUES only as a default.

primap2/pm2io/_data_reading.py
```diff
@@ -191,37 +219,45 @@ def convert_long_dataframe_if(
...
+        data_copy, coords_cols, add_coords_cols, coords_defaults, coords_terminologies
     )
-
     attrs.update(naming_attrs)
 
+    # replace the NA_Values
+    if convert_nan:
+        repl_dict = dict(zip(NA_VALUES, list(np.full(len(NA_VALUES), np.nan))))
+        data_copy["data"] = data_copy["data"].replace(repl_dict)
+        data_copy["data"] = pd.to_numeric(data_copy["data"], errors="coerce")
```
mikapfl, line 240: duplicated in `long` and `wide`, I moved it to a separate function.

mikapfl (changes requested):
Looks pretty good in general. I made some smaller changes for code standards and readability and have found some things that probably should be changed.

primap2/pm2io/_conversion.py
```diff
@@ -62,6 +63,11 @@ def convert_unit_primap_to_primap2(unit: str, entity: str) -> str:
     # remove spaces for more flexibility in input units
     unit = unit.replace(" ", "")
 
+    # check if entity contains GWP information. If so discard
+    # not needed for PRIMAP1 entities but when using the function for data reading
```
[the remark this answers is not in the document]
JGuetschow: I don't like updating the name because it means changing code that uses the function (though it's not too much code). But it's also weird to have a function that does more general conversion and has primap1 as a special case while still having a primap1 conversion function. It could be a wrapper that does a format check and then calls the general function.

primap2/pm2io/_data_reading.py
```diff
@@ -38,6 +40,27 @@
...
+    "C",
+    "NaN",
+    "IE",  # this is a values that should be mapped to 0
+    "IE, NE, NO",
+    "IE,NA,NO",
+    "NA,IE,NO",
+    "NO,NE,IE",
+    "NE, NO",
+]
+# TODO: distiguish between values mapped to NA and values mapped to 0, e.g. IE which
```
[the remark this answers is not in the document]
JGuetschow: Well, it's how it's done in primap1. But of course it's much better to do it right now. But I think it needs some discussion. So maybe we leave it as a TODO here and open an issue for it?

primap2/pm2io/_data_reading.py
```diff
@@ -191,37 +219,45 @@ def convert_long_dataframe_if(
     if add_coords_cols:
         check_overlapping_specifications_add_cols(coords_cols, add_coords_cols)
 
-    filter_data(data_long, filter_keep, filter_remove)
+    # make a copy to keep input dataframe unchanged
+    data_copy = data_long.copy(deep=True)
```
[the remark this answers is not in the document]
JGuetschow: OK, sounds good to use a parameter. I don't really like functions that alter your input to something undefined, though. But maybe it's OK if the docs mention it prominently.

primap2/pm2io/_data_reading.py
```diff
@@ -433,6 +469,7 @@ def convert_wide_dataframe_if(
     meta_data: Optional[Dict[str, Any]] = None,
     time_format: str = "%Y",
     time_cols: Optional[List] = None,
+    convert_nan: Optional[bool] = False,
```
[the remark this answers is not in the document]
JGuetschow: So I could pass `None` as parameter? That means that whenever you do not allow `None~ you should not use `Optional`?

primap2/pm2io/_data_reading.py
JGuetschow: Yes, that looks unfinished. It should raise the error and give the information on the strings that caused it.

primap2/pm2io/_data_reading.py
JGuetschow: Good. Splitting functions is something I usually forget.

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
JGuetschow: Based on the function to replace NA values it seems to be very easy to implement once we have decided what to map to 0 and what to NA. So maybe we should implement it right away after all.

primap2/pm2io/_GHG_inventory_reading.py
JGuetschow: OK, can be done. But I'll curse you when I have to adapt all BUR reading scripts.

primap2/pm2io/_conversion.py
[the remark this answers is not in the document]
mikapfl: We could use a new name, and keep the old name for compatibility. Doesn't cost anything to have
```python
convert_unit_primap_to_primap2 = convert_unit_to_primap2
```
essentially forever.

primap2/pm2io/_data_reading.py
JGuetschow: A user defined list would be best. Especially with the option to define what is mapped to 0 and what is mapped to NA (and a default conversion listed explicitly in the docs).

primap2/pm2io/_data_reading.py
JGuetschow: OK, great

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
mikapfl: Yeah, it is true that it is not nice to eat parameters. We don't in public primap2 API, to honor the principle of least surprise - I think we originally designed `convert_long_dataframe_if` as private API or something?

But I think a parameter with which we can change the safe default to a fast non-default would be fine. The user would explicitly ask for it, with something like `consume_input=True` or so.

JGuetschow:
There now are quite a few things to implement. I'll work on them and will amend the pull request accordingly.

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
JGuetschow: OK, so default copy and optionally mess up input but increase speed

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
mikapfl: I think `IE` and `NO` (and their combinations) are quite obviously 0, while `NE` and `nan` are obviously NaN. Everything else is more difficult. `NA` (not applicable) is not necessarily obviously NaN, because `NA` often means something like "does not apply in this category". So, I think `NA` combined with `IE` and/or `NO` needs to be 0 as well. `NA,NE` on the other hand should be NaN. About a lonely `NA` I'm not so sure. Maybe NaN is the safer choice there?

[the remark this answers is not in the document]
JGuetschow: Based on these definitions
* NO: not occurring 
* NE: not estimated 
* NA: not applicable
* IE: included elsewhere
* C: confidential
 
I agree with you base mapping and would map NA to NaN: 

* NO: 0
* NE: NaN
* NA: NaN
* IE: 0
* C: NaN (if it's confidential there might be data from other sources we can use)
* -: 0, as it's often used as a substitute for 0 in Excel

For NA, I choose NaN as I would say that not applicable is something based on the reporting guidelines not on the country circumstances. So it might be that there are emissions, but they are not reported under the guidelines. If on the other hand there are just no emissions then this should be the case for all datasets and we're fine with NaN.

Now for the combinations:  

* When IE is included I would use 0
* When NO is included I would use 0 as well
* NA, NE is mapped to NaN as both are mapped to NaN individually

In total we have the following mapping for all the values that occurred so far. We could also use regular expressions to implement the rules above to cover new cases instead of adding to the list all the time.

```
NA_VALUES = {
    "IE": 0,  
    "NE": np.nan,
    "NE0": np.nan,
    "NO": 0,
    "C": np.nan,
    "NaN": np.nan,
    "nan": np.nan,
    "-": 0,
    "NA, NE": np,nan,
    "NA,NE": np.nan,
    "NO,NE": 0, 
    "NO, NE": 0,   
    "NE,NO": 0,
    "NE, NO": 0,
    "IE, NE, NO": 0,
    "IE,NA,NO": 0,
    "NA,IE,NO": 0,
    "NO,NE,IE": 0,
}
```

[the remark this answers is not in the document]
mikapfl: sounds good!

A list is fine, I think, but we could also instead use a function:

```python

_special_codes = {
    'C': np.nan,
    'NaN': np.nan,
    'nan': np.nan,
    '-': 0,
    'NE0': 0
}

def parse_code(code: str) -> float:
    code = code.strip()
    if code in _special_codes:
        return _special_codes[code]

    parts = code.split(',')
    parts = [x.replace('.', '').strip() for x in parts]
    if 'IE' in parts or 'NO' in parts:
        return 0
    if 'NE' in parts or 'NA' in parts:
        return np.nan
    raise ValueError(f'Could not parse code: {code!r}.')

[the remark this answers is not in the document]
Comment (mikapfl) [reply primap2/pm2io/_data_reading.py:53]: The advantage is that stuff like `IE,1` where the "1" is a footnote will also be parsed properly and added dots and spaces are handled generally, so it might work for more things. The disadvantage could be that a function is slower and initially harder to read.

File: primap2/pm2io/_conversion.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (JGuetschow) [reply primap2/pm2io/_conversion.py:66]: I renamed the function without adding an alias. We don't have enough usage yet to justify adding potentially confusing aliases.

File: primap2/pm2io/_data_reading.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (JGuetschow) [reply primap2/pm2io/_data_reading.py]: Implemented and just pushed the code.

File: primap2/pm2io/_data_reading.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (JGuetschow) [reply primap2/pm2io/_data_reading.py:53]: The function is nice as it will also catch a lot of cases not on our current list. 

Maybe we could give the user the option to specify a dict for a mapping and we use the function as the default. This would be handy if one just wants to change individual mappings. On the other hand a strict distinction between user defined mappings and internal mappings would also be useful for clarity for the users which means either internal mappings or user defined mappings. (see also the convert_nan discussion below)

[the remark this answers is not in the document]
Comment (mikapfl) [reply primap2/pm2io/_data_reading.py:53]: > Maybe we could give the user the option to specify a dict for a mapping and we use the function as the default.

Sounds good.

I didn't understand your "on the other hand". What is the alternative proposal?

File: primap2/pm2io/_conversion.py [hunk 10]
```diff
@@ -18,11 +18,12 @@
 
 
 def convert_unit_primap_to_primap2(unit: str, entity: str) -> str:
```
mikapfl, line 20: You didn't rename the function, though.

JGuetschow: There was another confirmation in the refactor dialogue that I missed.

primap2/pm2io/_GHG_inventory_reading.py
JGuetschow: Now implemented

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
JGuetschow: First option: user defined rules, after these the function is used for the rest. 
Second option: when the user defines rules only these are used and not the function.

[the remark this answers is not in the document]
mikapfl: I would prefer the second option. If a user wants to change things, they get to define everything themselves.

primap2/pm2io/_data_reading.py
JGuetschow: I've added code to re-raise the error. This will not happen when using the official primap2 csv reading functions though, as they already remove the strings

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
JGuetschow: A small addition: I would map just spaces to np.nan because it's likely an empty cell read incorrectly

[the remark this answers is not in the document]
JGuetschow: Do you map NE0 to 0 instead of nan on purpose? It's mapped to nan in the list solution but to 0 in the function solution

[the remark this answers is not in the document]
mikapfl: No, probably a mistake because I have no idea what `NE0` means.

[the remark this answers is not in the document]
JGuetschow: I've now implemented the changes. Str conversion is controllable from the outside. The function is translated into a dict for all values actually occurring in the DataFrame. I've pushed the changes. More tests and docs to come. The behavior of the internal read_csv functions has changed. They now do not convert any strs to NaN but return the raw csv data.

primap2/pm2io/_data_reading.py
JGuetschow: See discussion of string conversion above. A user controllable conversion has been implemented now.

primap2/pm2io/_data_reading.py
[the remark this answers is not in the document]
JGuetschow: The functions has now been implemented and all data reading functions have been changed to use it (or a user specified conversion). The "read_csv" functions now do not do any string processing but return the raw data. All string processing is done in the "convert_XXX_dataframe_if" functions.

JGuetschow:
All changes have now been implemented and test have been added. The copy_df functionality is not tested as the conversion functions are not tested individually and our data reading functions do not use the copy_df functionality.

I will go through the docs and see if changes and additions are needed.

primap2/pm2io/_data_reading.py
```diff
@@ -191,37 +205,55 @@ def convert_long_dataframe_if(
...
+
+    if isinstance(convert_str, dict):
+        # get data columns (just one as we have long format)
+        data_cols = ["data"]
+        # find all string values
+        str_values = find_str_values_in_data(data_copy, data_cols)
+        # create replacement dict
+        str_repl_dict = create_na_replacement_dict(str_values, convert_str)
+        replace_na_values(data_copy, data_cols, str_repl_dict)
+
```
mikapfl, line 237: The function silently does nothing if I e.g. pass a string for `convert_str`. Would be better to raise a ValueError. Maybe the whole type handling can be shifted into `create_na_replacement_dict`? I could make a patch if you concur that an error should be raised.

primap2/pm2io/_data_reading.py
```diff
@@ -1244,17 +1310,150 @@ def rename_columns(
...
+def find_str_values_in_data(
+    data: pd.DataFrame,
+    columns: List[str],
+) -> list:
+    """Find all string values occurring in given columns of a DataFrame"""
+    # limit our analysis to columns that contain strings
+    # (or other object types)
+    cols_with_strs = (
+        data[columns].select_dtypes(include=[object]).columns.values.tolist()
+    )
```
mikapfl, line 1340: What if a column is entirely string? Then it could also be of type `StringDtype`, right?

primap2/pm2io/_data_reading.py
```diff
@@ -1244,17 +1310,150 @@ def rename_columns(
...
+    temp = []
+    for col in cols_with_strs:
+        temp += list(data[col].unique())
+    temp = list(set(temp))
+    strs = [x for x in temp if not is_float(x)]
```
mikapfl, line 1345: `is_float("nan") == True` and therefore the string "nan" (and "NaN" and possible others) will never be included in the output of `find_str_values_in_data`. Does that cause problems? At least it is then unnecessary to include `"nan"` in `_special_codes`.

primap2/pm2io/_data_reading.py
```diff
@@ -1244,17 +1310,150 @@ def rename_columns(
...
+        if str_val in user_na_conv:
+            mapping[str_val] = user_na_conv[str_val]
+        else:
+            mapping[str_val] = parse_code(str_val)
+
+    return mapping
+
+
+def replace_na_values(data: pd.DataFrame, columns: List[str], na_repl_dict):
+    """Replace str values indicating not-a-number by float NaN."""
```
mikapfl, line 1383: It doesn't strictly use NA values, right? Default rules also have 0, and user rules could map stuff to 1.5 if they wanted.

mikapfl: So maybe `replace_values` would be a better name to avoid confusion?

primap2/tests/test_data_reading.py
```diff
@@ -46,6 +46,82 @@ def test_metadata_for_variable(unit, entity, expected_attrs):
...
+        (
+            "25,000",
+            False,
+        ),  # be careful when reading data to process thousands seperators
+        ("25,00", False),
+        ("IE, NO", False),
+    ],
+)
+def test_is_float(to_test_for_float, expected_result):
+    assert pm2io._data_reading.is_float(to_test_for_float) == expected_result
```
mikapfl, line 64: This is test is nice, just as inspiration: it is also possible to do doctests like I did in the docstring of `preferred_unit`. They are also automatically run etc. Sometimes easier to do since you can define them directly in the docstring of the function instead of making a special test function in another file. For simple input -> output tests doctests work great, for more complex tests a separate function is definitely better.

mikapfl (approved):
Looks really good! You can merge when you are ready, I just added some things I noticed

primap2/pm2io/_data_reading.py
JGuetschow: I thought pandas always uses object when the data contains strings.

primap2/pm2io/_data_reading.py
JGuetschow: right

primap2/pm2io/_data_reading.py
JGuetschow: I wondered why nan and NaN  were in the original list but didn't really think about it more and kept them. As `replace_values` does an explicit type conversion we don't need to replace them anyway as they will be converted fine. The explicit type conversion is necessary because without it we might get columns with dtype `int`. So I'll remove them from the list.

primap2/pm2io/_data_reading.py
JGuetschow: I agree that raising an error makes sense. I've changed it, push will come soon

primap2/pm2io/_data_reading.py
JGuetschow: Actually they need to be in the list, else parse code fails.

primap2/pm2io/_conversion.py
```diff
@@ -157,70 +159,145 @@ def convert_ipcc_code_primap_to_primap2(code: str) -> str:
...
+        "M3D2LU": "M.3.D.2.LU",
+        "M.AG": "M.AG",
+        "M.AG.ELV": "M.AG.ELV",
+        "M.BK": "M.BK",
+        "M.BK.A": "M.BK.A",
+        "M.BK.M": "M.BK.M",
+        "M.LULUCF": "M.LULUCF",
+        "M.MULTIOP": "M.MULTIOP",
+        "M.0.EL": "M.0.EL",
+        "M.BIO": "M.BIO",
```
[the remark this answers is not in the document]
mikapfl: nah, too complicated.

PR #117, opened 2022-07-26
Fixes for new xarray version, merging

mikapfl:
# Pull request

- [x] Tests added

## Description

Tests for merging were broken with the new xarray version. To unbrake everything, I had to refactor the merging because I was honestly not able to understand what was going on in that super large function.


mikapfl:
@JGuetschow Maybe you want to run some tests on this new version, I changed for example the log messages in the refactoring - I'm not sure if you somehow rely on the exact log message format.

In the refactoring, I also removed the conversion to pandas dataframes and do the tolerance checking natively in xarray instead. I *think* via xarray we could also avoid the recursive for-loops to get to timeseries before doing the comparison. That could speed up merging *a lot*. Do I miss some reason why the merging algo is recursive?

PR #118, opened 2022-07-27
Faster merge

mikapfl:
# Pull request

- [x] Tests added

## Description

Use xarray's `DataArray.combine_first` function to do the heavy lifting in `pr.merge` instead of our own recursive algorithm. This is at least 10x faster than the previous algorithm for our tests, and easier to understand.

### DIFF
--- a/changelog_unreleased/faster-merge.rst
+++ b/changelog_unreleased/faster-merge.rst
@@ -1,0 +1,1 @@
+* pr.merge: use xarray's combine_first instead of own algorithm for better performance.
--- a/primap2/_merge.py
+++ b/primap2/_merge.py
@@ -1,8 +1,8 @@
 """Merge arrays and datasets with optional tolerances."""
 
 import typing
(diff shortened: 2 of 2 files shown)


mikapfl:
@JGuetschow do you know if there are any issues still blocking this? It is quite a large performance enhancement, so I'd like to merge it, but also don't want to break your workflows.

JGuetschow:
I don't know of any issues, but so far I've not looked at it, just overlooked the notification. Sorry for that. 
Your strategy is to split discrepancy detection and merging, correct. You first calculate the differences and throw an error / warning based on that and later just merge with combine_first. This should work fine.

mikapfl:
Yeah, and I also changed the format of the warnings. I don't know if you are already using the warnings downstream and expect a specific format.

JGuetschow:
No, I'm not expecting a specific format (an just realized that the entity was missing in my output)

PR #121, opened 2022-08-15
Data reading fixes

JGuetschow:
# Pull request

- [ ] Tests added

## Description

Adds small fixes to data reading functions. The main fix is that if data exists in both GWP equivalents and and native units for an entity the GWP equivalent data is converted to native units. This was not done before and made conversion to native primap2 format impossible.
I have not added a test so far as this is a quite special issue.


mikapfl:
> This was not done before and made conversion to native primap2 format impossible.

How so? The `minimal_ds` example has SF6 in GWP and native units:

```python
In [1]: import primap2 as pm2

In [2]: import primap2.tests

In [3]: pm2.tests.examples.minimal_ds()
Out[3]: 
<xarray.Dataset>
Dimensions:          (time: 21, area (ISO3): 4, source: 1)
Coordinates:
  * time             (time) datetime64[ns] 2000-01-01 2001-01-01 ... 2020-01-01
  * area (ISO3)      (area (ISO3)) <U3 'COL' 'ARG' 'MEX' 'BOL'
  * source           (source) <U8 'RAND2020'
Data variables:
    CO2              (time, area (ISO3), source) float64 [CO2·Gg/a] 0.5118 .....
    SF6              (time, area (ISO3), source) float64 [Gg·SF6/a] 0.8313 .....
    CH4              (time, area (ISO3), source) float64 [CH4·Gg/a] 0.2012 .....
    SF6 (SARGWP100)  (time, area (ISO3), source) float64 [CO2·Gg/a] 1.987e+04...
Attributes:
    area:     area (ISO3)
```

Or do I misunderstand what the aim is here?

JGuetschow:
I think the problem is that the GWP units were converted before but not to the native units already existing but to default units. This resulted in two different native units and that is not allowed in the pm2 native format. 
Sorry for the bad description. I made this fix a few weeks ago but failed to create the pull request before being off for four weeks.

primap2/_merge.py
```diff
@@ -54,8 +54,10 @@ def merge_with_tolerance_core(
     da_merge: xr.DataArray
         data to merge into ds_start
     tolerance: float (optional), default = 0.01
-        The tolerance to use when comparing data
-    error_on_discrepancy: (optional), default = True
+        The tolerance to use when comparing data. Tolerance is relative to values in
+        the calling Dataset. Thus by default a 1% deviation of values in da_merge
+        from the calling Dataset is tolerated.
+    error_on_discrepancy (optional), default = True
```
mikapfl, line 60: Hm, this doesn't seem to have anything to do with data reading? Is this some merge conflict with #118 ?

primap2/pm2io/_data_reading.py
```diff
@@ -1394,6 +1394,9 @@ def preferred_unit(entity: str, units: List[str], gwp_to_use: Optional[str]) ->
 
     In general, "Gg <substance> / year" will be preferred if it is compatible with the
     given input units. Otherwise, the first unit from units will be preferred.
+    If gwp_to_use is not None native units without GWP will be preferred. If it is
+    not set but conversion without gwp context is impossible Gg <unit_substance> / yr
+    will be returend as preferred unit if the first unit from units can be converted.
```
mikapfl, line 1399: Isn't `Gg <unit_substance> / yr` a native unit? If so, then I don't understand the docs. As far as I understand: If `gwp_to_use` is `None` ("not set"), then native units should be preferred, and then if a GWP context is not available, things are weird.

primap2/pm2io/_data_reading.py
```diff
@@ -1418,7 +1421,7 @@ def preferred_unit(entity: str, units: List[str], gwp_to_use: Optional[str]) ->
     >>> preferred_unit("CH4", ["kt CO2 / yr"], "AR4GWP100")
     'Gg CH4 / yr'
     >>> preferred_unit("CH4", ["kt CO2 / yr", "Mg CO2 / yr"], None)
-    'kt CO2 / yr'
+    'Gg CO2 / yr'
```
mikapfl, line 1424: But why should this be a better result than `kt CO2 / yr`? That seems very specialized to always prefer `Gg`, even if the input data are not ever given in Gg.

primap2/tests/test_data_reading.py
```diff
@@ -122,6 +122,21 @@ def test_create_na_replacement_dict(strs, user_na_conv, expected_result):
...
+        ("CH4", ["kt CO2 / yr", "Mg CO2 / yr"], None, "Gg CO2 / yr"),
+    ],
+)
+def test_preferred_unit(to_test_entity, to_test_units, to_test_gwp, expected_result):
+    assert (
+        pm2io._data_reading.preferred_unit(to_test_entity, to_test_units, to_test_gwp)
+        == expected_result
+    )
+
+
```
mikapfl, line 139: Don't duplicate doctests. Those are already as testable (and tested) examples in the docstring, duplicating them here will just add more work down the line.

mikapfl (commented):
Hi,

in general, I think we should not do this on import. If the same stuff is given in GWP and without, we should read in two different variables into the primap2 dataframe. If the user then wants to convert that and fill gaps from one variable with data from the other, or whatever, they can do that easily and explicitly using `pr.convert_to_mass()`/`pr.convert_to_gwp()` and `pr.set()`. We shouldn't convert the data already when importing it.

primap2/pm2io/_data_reading.py
mikapfl: Ah, maybe the only info missing there is this:
> If gwp_to_use is None, and in `units` are GWP units as well as native units and conversion between them without gwp context is impossible, `Gg <unit_substance> / yr` will be returned as preferred unit if at least the first unit from `units` can be converted.

That way, I understand *what* it is doing (but not why).

primap2/_merge.py
JGuetschow: Yes, that edit was due to a merge conflict. I don't know why it shows up here, because I thought I had changed it to the head version from my local version.

primap2/pm2io/_data_reading.py
JGuetschow: Maybe Gg was the standard in primap1 and I got used to it. It's important however, that a default unit is given that is both used when converting from GWP weighted units and native units because else we will end up with different native units for the same entity. A way around this would be to convert the GWP weighted units to the present native unit in case it is present. Don't know how much work this is though.

primap2/tests/test_data_reading.py
JGuetschow: Sorry, I overlooked the doctests.

primap2/pm2io/_data_reading.py
```diff
@@ -1434,8 +1437,15 @@ def preferred_unit(entity: str, units: List[str], gwp_to_use: Optional[str]) ->
             ureg[native_unit], *conversion_contexts
         ):
             return native_unit
+        else:
+            # try to convert to Gg of the substance present in the unit
+            # extract the substance from the fallback unit
+            substance = unit_fallback.split(" ")[1]
```
mikapfl, line 1443: This is not robust. There is no need that the input will be of the form "<mass> <substance> / <time>", it could be also of the form "<substance> <mass> / <time>" or any other number of things. I think if you really need to go down this rabbit hole, you should just assume `gwp_unit = "Gg CO2 / yr"` and try that (note that please don't re-use the `native_unit` variable here, because it definitely is *not* the native unit, the native unit is in this context the mass unit, not the gwp unit (see above, where `native_unit = "Gg " + entity + " / yr"`).

primap2/pm2io/_data_reading.py
JGuetschow: To convert to the first non-GWP unit instead of Gg the harmonize_units function needs a structural change. Currently it iterates over all entities. The entities for GWP weighted units and native units are different as the GWP is included in the entity name (e.g. "CH4 (SARGWP100)"). Thus to change it we would have to first group the entities by "basic entity" and then do the conversion. I can implement that but not in the next weeks as I'm busy with the PRIMAP-hist update

primap2/pm2io/_data_reading.py
```diff
@@ -1506,9 +1516,9 @@ def harmonize_units(
         data_this_entity = data.loc[data[entity_col] == entity]
         units_this_entity = data_this_entity[unit_col].unique()
 
-        if len(units_this_entity) > 1 or gwp_to_use:
+        if len(units_this_entity) > 1:
```
mikapfl, line 1519: What happens if `gwp_to_use` is set, but `len(units_this_entity) = 1`? I.e. I have a GWP entity in a single unit. Previously, I would get unit conversion, now not any more. Is that intended? I see that we don't really need unit conversion in that case, as things are harmonious already, just checking.

primap2/_merge.py
mikapfl: probably some weirdness, anyway, whatever is new in your commit is clearly better than what was before, so let's take that. :slightly_smiling_face:

primap2/pm2io/_data_reading.py
mikapfl: What I don't understand is: Why are we changing the entity in `harmonize_units` at all? If we iterate over entities and treat GWP entities and mass entities as different entities, then we will never have a problem if they have non-matching units, right?

JGuetschow:
> Hi,
> 
> in general, I think we should not do this on import. If the same stuff is given in GWP and without, we should read in two different variables into the primap2 dataframe. If the user then wants to convert that and fill gaps from one variable with data from the other, or whatever, they can do that easily and explicitly using `pr.convert_to_mass()`/`pr.convert_to_gwp()` and `pr.set()`. We shouldn't convert the data already when importing it.

So far we did convert on import when possible. I think it makes sense as we have proper unit handling in PRIMAP2 and the data information is the same. But your point also makes sense and in case of errors in the data or unclear GWP information it might be harder to find them if we do the conversion on import.

I think we can then discard this pull request (or save it for later with modifications) and I'll make a new one (or update this one) once I've changed the import functions. I've actually started a function to unify the GWPs which could then be used by the user, but it's not done so far.

primap2/pm2io/_data_reading.py
JGuetschow: True. That probably comes from thinking only about my use cases where I use a defined format.

JGuetschow:
I see that the reason this was not yet merged is not that I just forgot to make the pull request, but that I have not checked everything yet (remembered that I had done that). So I will come back later when I have solved all the issues.

PR #122, opened 2022-08-23
fix in to_interchange_format

JGuetschow:
# Pull request

- [ ] Tests added
No new functionality added, tests  are not affected. One could add a tests for the bug that has been fixed.

## Description

Fix in to_interchnage_format. Before timeseries which contained only nan were kept (I assume the intention was to remove them, as there was a dropna in the code). On the other hand time series which contained only a few (or single) data points were in some cases dropped. This has now been fixed by specifying the time columns as subset in dropna.

A further tiny fix is to make regexp_unit optional in nir_add_unit_information as was intended already before, but not indicated in the function header.


changelog_unreleased/fix-data-reading.rst
```diff
@@ -0,0 +1 @@
+* Fix handling of units when reading data where CO2eq and native units coexist
```
mikapfl, line 1: ? I don't see any changes in the code related to this.

primap2/_merge.py
```diff
@@ -54,8 +54,10 @@ def merge_with_tolerance_core(
     da_merge: xr.DataArray
         data to merge into ds_start
     tolerance: float (optional), default = 0.01
-        The tolerance to use when comparing data
-    error_on_discrepancy: (optional), default = True
+        The tolerance to use when comparing data. Tolerance is relative to values in
+        the calling Dataset. Thus by default a 1% deviation of values in da_merge
+        from the calling Dataset is tolerated.
+    error_on_discrepancy (optional), default = True
```
mikapfl, line 60:
```suggestion
    error_on_discrepancy: (optional), default = True
```
Why remove the `:`?

primap2/_merge.py
```diff
@@ -253,8 +255,10 @@ def merge(
         da_merge: xr.DataArray
             data to merge to the calling object
         tolerance: float (optional), default = 0.01
-            The tolerance to use when comparing data
-        error_on_discrepancy: (optional), default = True
+            The tolerance to use when comparing data. Tolerance is relative to values in
+            the calling Dataset. Thus by default a 1% deviation of values in da_merge
+            from the calling Dataset is tolerated.
+        error_on_discrepancy (optional), default = True
```
mikapfl, line 261:
```suggestion
        error_on_discrepancy: (optional), default = True
```

mikapfl (changes requested):
Thanks for finding the problem + fixing it. I'd appreciate a test, since I don't fully understand what is going on and why `how = "all"` doesn't do what I thought it would. With a test case, we could make sure we're not getting regressions later on. But it might be difficult to engineer a nice test case.

Other changes that I requested are just typos I think and an inadvertently included unrelated change.

JGuetschow:
I will add a test. I was also puzzled that "all" didn't do what it supposedly should. Maybe it's a bug. When we have a minimal test we can investigate further.

mikapfl:
I think we can merge the fix also without the minimal test case. Better have the fix without a test than to have an unfixed primap.

JGuetschow:
I think my description is wrong. Only one of the two issues was present, namely keeping the all-nan timeseries. The dropping of singe data points was a mistake by me, I think. Thus I think merging is fine without a test, because the fixed issue is actually small. The larger issue did not exist in the first place, I think.

Issue #206, opened 2024-04-02
Setup allows installation of unsupported Python version 3.12 


crdanielbusch:
**Describe the bug**

Python 3.12 is explicitly not supported, but is not excluded in the setup configuration.

**Expected behavior**

The setup should only allow the installation of supported Python versions.

**Additional context**

A problem we discovered with Python version 3.12 is described in https://github.com/pik-primap/primap2/issues/205#issue-2220147878

PR #207, opened 2024-04-02
Tighten dependencies on python versions

mikapfl:
# Pull request

- [ ] Tests added

## Description

Specify minimal and maximal python versions in our python version dependencies.

I guess the `.check_python_version.py` script is not necessary any more (pip rejects this by itself using the configuration in `setup.cfg`), but it has the advantage that it is run in Makefile before anything else when setting up a development environment, so it provides a fast feedback and readable error message. So, I just left it in.

Closes: #206

### DIFF
--- a/.check_python_version.py
+++ b/.check_python_version.py
@@ -1,14 +1,23 @@
 """Check if the used version of Python is good enough for us."""
 
-# ruff: noqa
+import itertools
 import sys
 
-MIN_VERSION = (3, 9)
+SUPPORTED_MAJOR_VERSIONS = (3,)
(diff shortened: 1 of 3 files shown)


mikapfl:
@crdanielbusch Could you check if this makes sense? I think it is rather obvious, but maybe I missed something. I tested that it now fails when trying to install on Python 3.12 on linux.

crdanielbusch:
It works on my Mac with Python 3.12 as well. So when I try to set up the virtual environment I will see an error message.

crdanielbusch:
It all looks good to me. Johannes and I excluded Python 3.12 in the same way in the UNFCCC_non_AnnexI repo.

PR #208, opened 2024-04-03
Compose: Use entities in priority and strategy selectors, refactor

mikapfl:
# Pull request

- [x] Tests added

## Description

* general cleaning up of `compose`
* strategy and priority selectors:
    - support and test multiple values per dimension (priority: not for priority dimensions)
    - support selecting by entity (not for priority dimensions)

### DIFF
--- a/primap2/csg/_compose.py
+++ b/primap2/csg/_compose.py
@@ -56,6 +56,8 @@ def compose(
         e.g., possible to define a different priority for a specific country by listing
         it early (i.e. with high priority) before the more general rules which should
         be applied for all other countries.
+        You can also specify the "entity" in the selection, which will limit the rule
+        to a specific entity (xarray data variable).
     strategy_definition
         Defines the filling strategies to be used when filling timeseries with other
         timeseries. Again, the priority is defined by a list of selections and
@@ -64,6 +66,8 @@ def compose(
         define a default strategy which should be used for all timeseries unless
         something else is configured, configure an empty selection as the last
         (rightmost) entry.
+        You can also specify the "entity" in the selection, which will limit the rule
+        to a specific entity (xarray data variable).
     progress_bar
         By default, show progress bars using the tqdm package during the
         operation. If None, don't show any progress bars. You can supply a class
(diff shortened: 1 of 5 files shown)


mikapfl:
Should be merged after #199 , also the changes view in github doesn't make a lot of sense currently because it compares to the main branch, will make more sense when #199 is merged.

primap2/csg/_compose.py
```diff
@@ -77,52 +78,52 @@ def compose(
...
             pbar = None
         else:
             pbar = progress_bar(total=number_of_timeseries, unit="ts", unit_scale=True)
+
         iterate_next_fixed_dimension(
             input_da=input_da,
-            priority_definition=priority_definition,
-            strategy_definition=strategy_definition,
+            priority_definition=priority_definition.limit("entity", entity),
+            strategy_definition=strategy_definition.limit("entity", entity),
```
mikapfl, line 126: it was surprisingly easy to enable filtering by entity, these two lines are all that's needed.

JGuetschow: Nice. didn't expect that but in hindsight I could have expected it as we're working on a per entity basis anyway, so no need for additional filtering of entities.

setup.cfg
```diff
@@ -77,6 +77,7 @@ dev =
     mypy
     tox
     numpydoc
+    ruff
```
JGuetschow, line 80: So for ruff is not run by default, right? I expect a lot of formatting changes and complaint about future deprecations when running ruff on the existing code. (We have to do it a some point ...)

JGuetschow (approved):

setup.cfg
mikapfl: I'm running ruff on the repo by default, it doesn't do any formatting changes anymore, they have been done already. They were minor (it is pretty close to black's style).

With the default ruleset, also `ruff check` is clean. I haven't checked with the CR ruleset.

mikapfl: Maybe this wasn't clear: We are running ruff already in the pre-commit (so also on github in the checks) in both checking and formatting mode. We have a reduced set of ruff check rules active compared to other CR projects. If you added the documentation rules and some other stuff, we would get lots of warnings, some of them actually things we should fix, some of them just stupid. At some point I could look through this, but I don't think it is pressing.

PR #213, opened 2024-04-09
Fix entity limits broken in strategy definition

mikapfl:


mikapfl:
Hm, I did this PR from pycharm, which works, but unfortunately doesn't use the PR templates which are configured in the repo, so it doesn't have the nice checklist. Well, was worth a try.

mikapfl:
# Pull request

- [x] Tests added

## Description

Fix a bug where the entity selection in strategy definitions is used in matching later on, which means nothing will match.

Note: It's hard to test this properly in the main branch because we only have one single strategy.

JGuetschow:
I'll wait for the CI before merging.

PR #215, opened 2024-04-09
ugly: add a hack to monkey-patch pint's ndim for speed

mikapfl:
# Pull request

- [ ] Tests added

## Description

monkey-patch the changes from https://github.com/hgrecco/pint/pull/1959 into pint so that performance goes back to normal

### DIFF
--- a/primap2/_units.py
+++ b/primap2/_units.py
@@ -13,6 +13,26 @@ from . import _accessor_base
 pint_xarray.setup_registry(ureg)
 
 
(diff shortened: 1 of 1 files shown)


mikapfl:
@JGuetschow it is not pretty at all, but the performance of `ndim` is unbearable. Everything is so super slow. This is also the reason why reading and writing netcdf files is slow. Hopefully, this gets resolved in pint proper, then we can remove this hack again.

JGuetschow (approved):
Nice

PR #223, opened 2024-04-19
Support skipping strategies, sources

mikapfl:
# Pull request

- [x] Tests added

## Description

Support:
* skipping strategies which are unable to process a specific timeseries, continuing with the next configured strategy.
* skipping specific input timeseries using `exclude_input` in the `PriorityDefinition`, useful e.g. when a specific source is wrong for a specific country.
* skipping output timeseries entirely using `exclude_result` in the `PriorityDefinition`, useful e.g. when a category is generally invalid for an entity.

Add:
* `primap2.Not` special value to use with `pr.loc` to select everything but a specified value or values along the specified dimension. 

Also:
* refactor CSG tests into a few more files, things were getting a bit confusing.


mikapfl:
Still a draft, tests are failing, the null strategy doesn't exist yet, tests aren't comprehensive yet.

mikapfl:
Still needed:
* [ ] review testing, add more tests if needed
* [ ] convert SkipStrategy() into real exported strategy, not local to classes

Because usage documentation outside the docstrings needs a more general documentation restructuring, I think it should be in a second PR to keep this PR readable and reasonably small.

mikapfl:
> convert SkipStrategy() into real exported strategy, not local to tests

Argh, the naming is confusing. SkipStrategy() as currently implemented should be named `ErrorStrategy` and is not useful outside of tests. We probably need a strategy which returns the input `ts` unchanged successfully, which can actually be used to skip a specific source for a specific timeseries without discarding input data.

primap2/csg/_models.py
```diff
@@ -136,10 +266,18 @@ class StrategyDefinition:
...
+    def find_strategies(
+        self, fill_ts: xr.DataArray
+    ) -> typing.Generator[FillingStrategyModel, None, None]:
+        """Yields all strategies to use for the timeseries, in configured order."""
         for selector, strategy in self.strategies:
-            if self.match(selector=selector, fill_ts=fill_ts):
-                return strategy
-        raise KeyError(f"No matching strategy found for {fill_ts.coords}")
+            if match_selector(selector=selector, ts=fill_ts):
+                yield strategy
```
mikapfl, line 280: We use the python Generator protocol here with a co-routine.

Co-routines are special functions which `yield` instead of `returning. When you call `next(f)` on a function `f`, the function is run until it hits the next `yield`, then it is suspended and the yielded value is the return value of the `next` call. Instead of calling `next` explicitly, you can also iterate, e.g. with a for loop, which then will always run the function up to the next `yield` whenever it needs the next value in the loop. Turns out, this is exactly what we need for strategies.

JGuetschow (approved):
Looks good to me. My larger example works with `exclude_result` instead of the `Null Strategy` and the result is the same.
Performance is a bit lower than before (roughly 10%). I've not done any profiling though.

mikapfl:
I'll look into the `entity` vs `variable` issue in a separate PR.

PR #226, opened 2024-04-26
Configuration files for compose

mikapfl:
# Pull request

- [ ] Tests added

## Description

Adds the ability to load `PriorityDefinition` and `StrategyDefinition` from configuration files. In particular, we use strict yaml files because they are type safe and reasonably easy to read.


mikapfl:
So far, only `PriorityDefinition` is implemented, `StrategyDefinition` will be a bit more work due to parsing the strategies and having to support custom strategies.

mikapfl:
We decided that adding configuration files with the features we want (definition of category lists, custom strategies, avoid repetition) adds too much complexity for too little benefit vs. python-based configuration at the current stage.

PR #227, opened 2024-04-29
CSG filling strategies and data aggregation functions

JGuetschow:
# Pull request

*Draft pull request. tests and docs will follow after initial discussion*

This PR is based on #223 which should be merged first

- [x] Tests added

## Description

* Least squares matching filling strategies
* functionality to aggregate along coordinates and entities
* function to add dimension including terminology in attrs


primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+
+import primap2
+
+
+@frozen
+class GlobalLSStrategy:
+    """Fill missing data by global least square matching.
+
+    The NaNs in the first timeseries `ts(t)` using harmonized data from the lower
+    priority timeseries `fill_ts(t)`. For harmonization we use
```
mikapfl, line 15:
```suggestion
    The NaNs in the first timeseries `ts(t)` is filled using harmonized data from
    the lower priority timeseries `fill_ts(t)`. For harmonization we use
```

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+    fill_ts(t)_h = ts(t) * a + b,
```
mikapfl, line 16:
```suggestion
    fill_ts(t)_h = fill_ts(t) * a + b,
```

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+    For the case `allow_shift = False` (b = 0) `scipi.optimize.least_squares` is used.
+
+    If there is no overlap in non-NaN data between ts(t) and fill_ts(t) the
+    `SubstitutionStrategy` is used.
+    """
+
+    allow_shift: bool = True
+    type = "globalLS"
+
+    def factor_mult(self, a, e, e_ref):
```
mikapfl, line 32:
```suggestion
    def _factor_mult(self, a, e, e_ref):
```
Really a very small thing, but if you don't want to make this function part of the public API of primap2, it should be marked private since the Strategy class will be part of the public API or primap2.

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+
+        If the class is initialized with `allow_shift = True` the faster
+        `GlobalLSlstsqStrategy` which is based on the `scipy.linalg.lstsq`
+        function is used.
+        For the case `allow_shift = False` (b = 0) `scipi.optimize.least_squares`
+        is used.
+
+        If there is no overlap in non-NaN data between ts(t) and fill_ts(t) the
+        `SubstitutionStrategy` is used.
+
```
mikapfl, line 63: Probably not very useful to repeat the documentation from the class docstring. Maybe:
```suggestion

    For a description of the algorithm, see the documentation of this class.

```

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+            data in ts is (partly) filled using scaled data from fill_ts.
+            descriptions contains information about which years were affected and
+            filled how.
+        """
+
+        filled_mask = ts.isnull() & ~fill_ts.isnull()
+        time_filled = filled_mask["time"][filled_mask].to_numpy()
+
+        if time_filled.any():
+            # check of we have overlap. if not use substitution strategy
```
mikapfl, line 89:
```suggestion
            # check if we have overlap. if not use substitution strategy
```

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+                            source=fill_ts_repr,
+                        )
+                    ]
+            else:
+                strategy = primap2.csg.SubstitutionStrategy()
+                filled_ts, descriptions = strategy.fill(
+                    ts=ts,
+                    fill_ts=fill_ts,
+                    fill_ts_repr=fill_ts_repr,
+                )
```
mikapfl, line 128:
```suggestion
            else:
                raise StrategyUnableToProcess(reason="No overlap between timeseries, can't match")
```

I think the choice what to do if Ls matching is not feasible should be left to the user.

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+
+class GlobalLSlstsqStrategy:
+    """
+    Fill missing data by global least square matching.
+
+    The NaNs in the first timeseries `ts(t)` using harmonized data from the
+    lower priority timeseries `fill_ts(t)`. For harmonization we use
+    fill_ts(t)_h = ts(t) * a + b,
+    where fill_ts(t)_h is the harmonized dataset and a and b are determined
+    by minimizing the least squares distance between ts(t) and fill_ts(t)_h.
```
mikapfl, line 153: Same typos as in the other strategy

primap2/csg/_strategies/global_least_squares.py
```diff
@@ -0,0 +1,267 @@
...
+                    description=f"no additional data in {fill_ts_repr}",
+                    function=self.type,
+                    source=fill_ts_repr,
+                )
+            ]
+
+        return filled_ts, descriptions
+
+
+class GlobalLSlstsqStrategy:
```
mikapfl, line 145: I'd convert this into a single function which just gets called by the other strategy. The user should never need to configure either one or the other strategy.

primap2/csg/_strategies/null.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+import xarray as xr
+
+import primap2
+
+
+class NullStrategy:
+    """return a nan time-series (used to skip certain fixed coordinate combinations)"""
```
mikapfl, line 8: I think this one is obsolete with `exclude_*`, no?

primap2/csg/_strategies/null.py
```diff
@@ -0,0 +1,54 @@
...
+            String representation of fill_ts. Human-readable short representation of
+            the fill_ts (e.g. the source).
+
+        Returns
+        -------
+            filled_ts, descriptions. filled_ts contains np.nan() only
+        """
+
+        filled_ts = ts
+        filled_ts.loc[:] = np.nan
```
mikapfl, line 44:
```suggestion
        filled_ts = ts.copy(deep=True)
        filled_ts.loc[:] = np.nan
```
otherwise you change data in `ts`, which is not allowed.

primap2/_aggregate.py
```diff
@@ -220,6 +220,99 @@ def fill_all_na(self, dim: Iterable[Hashable] | str, value=0) -> xr.DataArray:
...
+                            data_agg = data_agg.assign_coords(
+                                coords={add_coord: (full_coord_name, [add_coord_value])}
+                            )
+                        else:
+                            raise ValueError(
+                                f"Additional coordinate {add_coord} specified but not "
+                                f"present in data"
+                            )
+
+                    da_out = da_out.pr.merge(data_agg, tolerance=rule_tolerance)
```
mikapfl, line 311: I'm wondering if it makes sense to add a `merge`-like mode to `pr.set`. In theory, the category aggregation sounds like something which should be a `pr.sum` and a `pr.set` and nothing else, but here you have to do quite a lot of set-up to turn `data_agg` into something which can be consumed by `pr.merge`.

primap2/_aggregate.py
```diff
@@ -600,3 +693,120 @@ def fill_na_gas_basket_from_contents(
...
+                            "gwp_context": basket.split(" ")[1][1:-1],
+                        }
+                        ds_out[basket] = ds_out.pr.gas_basket_contents_sum(
+                            basket=basket,
+                            basket_contents=basket_contents_present,
+                            min_count=1,
+                        )
+                        entities_present.add(basket)
+                    except Exception as ex:
+                        print(f"No gas basket created for {basket}: {ex}")
```
mikapfl, line 811: At a minimum, you have to replace this with `logger.warning` or something, not `print` it.

primap2/_aggregate.py
```diff
@@ -600,3 +693,120 @@ def fill_na_gas_basket_from_contents(
...
+                    )
+                else:
+                    try:
+                        ds_out[basket] = xr.full_like(
+                            ds_out[basket_contents_present[0]], np.nan
+                        ).pr.quantify(units="Gg CO2 / year")
+                        ds_out[basket].attrs = {
+                            "entity": basket.split(" ")[0],
+                            "gwp_context": basket.split(" ")[1][1:-1],
+                        }
```
mikapfl, line 803: Maybe we could also add this functionality to `gas_basket_contents_sum` directly, that it can also operate if the basket doesn't exist, yet. That way, it is easy to add a single gas basket with one function call.

mikapfl (commented):
Nice, generally a good direction. I have some design ideas and the null strategy changes `ts`, which is not allowed (I added the new strategies to the simple protocol conformance test, which shows the problem).

primap2/_aggregate.py
JGuetschow: Had planned to do that and forgot.

primap2/_aggregate.py
JGuetschow: Yes, that would make things easier. But I think for this case I could also just use my `ds.pr.add_dim` function instead of doing the same steps manually

primap2/csg/_strategies/global_least_squares.py
JGuetschow:
```suggestion
    The NaNs in the first timeseries `ts(t)` are filled using harmonized data from the lower
    priority timeseries `fill_ts(t)`. For harmonization we use
```

primap2/csg/_strategies/global_least_squares.py
JGuetschow: Would the user have to add another item to the strategy list? As we would have to do that for every source / scenario combination in PRIMAP-hist individually this would make the strategy list very long with repetitive information. Or do we have a different way to specify this?
We could also add a fallback parameter to the strategy

primap2/csg/_strategies/global_least_squares.py
JGuetschow: Makes sense for the current code. I kept it as an individual strategy because I thought maybe I'll add a faster version for the b=0 case to this strategy.

primap2/csg/_strategies/null.py
JGuetschow: I think so. I forgot to remove it before the pr.

primap2/csg/_strategies/global_least_squares.py
mikapfl: You can simply have a fallback

```python
strategy_definition = StrategyDefinition([
    # For EDGAR and ANDREWS, shifting doesn't work well 
    ({"source": ["EDGAR", "ANDREWS"]}, GlobalLSStrategy(allow_shift=False)),
    # All other sources work well with shifting
    ({"source": ["CRF2021", "CRF2022", "CRF2023"]}, GlobalLSStrategy(allow_shift=True)),
    # If the least-squares strategy fails, use simple substitution instead
    ({}, SubstitutionStrategy())
])
```
Of course, using the default selector like this means if you forget to include a strategy in the sources list for both GlobalLSStrategies, it will just use `SubstitutionStrategy` instead of raising an error. To guard against this and maybe be a bit more explicit you could write it like this:
```python
primap_hist_sources = ["EDGAR", "ANDREWS", "CRF2021", "CRF2022", "CRF2023"]
no_shift_sources = ["EDGAR", "ANDREWS"]

strategy_definition = StrategyDefinition([
    # For EDGAR and ANDREWS, shifting doesn't work well 
    ({"source": no_shift_sources}, GlobalLSStrategy(allow_shift=False)),
    # All other sources work well with shifting
    ({"source": [source for source in primap_hist_sources if source not in no_shift_sources]}, GlobalLSStrategy(allow_shift=True)),
    # If the least-squares strategy fails, use simple substitution instead
    ({"source": primap_hist_sources}, SubstitutionStrategy())
])
```
Or are the rules much more complex than I imagine?

primap2/_aggregate.py
mikapfl: Cool, I'll see if I can add it to `gas_basket_contents_sum` easily, I'll make a PR against this branch.

primap2/csg/_strategies/global_least_squares.py
JGuetschow: I'll remove all internal fallbacks (to be consistent also the fallback which is triggered when negative data are created) and will deal with them through configuration. It might make debugging harder because a fallback might also be a default, but with a more explicit configuration I think we can deal with that.

docs/usage.ipynb
```diff
@@ -76,19 +95,487 @@
...
...r><br><br>    [[[[[0.8387044636461196 0.06611289471634418]]]<br><br><br>      [[[0.2969031501505053 0.41277875944623843]]]]<br><br><br><br>     [[[[0.25411629586572826 0.45533166019406457]]]<br><br><br>      [[[0.46922087458989514 0.7452963998006973]]]]]<br><br><br><br><br>    [[[[[0.543941930055117 0.3924759230683955]]]<br><br><br>      [[[0.6672538764869582 0.45849306759132824]]]]<br><br><br><br>     [[[[0.24802237609607003 0.8478432356756823]]]<br><br><br>      [[[0.7689461445858776 0.2473680862883174]]]]]]<br><br><br><br><br><br>   [[[[[[0.8652087348177219 0.7311187418393271]]]<br><br><br>      [[[0.3721644683017781 0.2667749819675661]]]]<br><br><br><br>     [[[[0.23671782297600097 0.9638031384997067]]]<br><br><br>      [[[0.16949888718549988 0.2942045586187396]]]]]<br><br><br><br><br>    [[[[[0.8337860461340107 0.6707973737352277]]]<br><br><br>      [[[0.02340306989276575 0.8311325800904432]]]]<br><br><br><br>     [[[[0.6086538800308976 0.17995373127512915]]]<br><br><br>      [[[0.6617146907776896 0.5068033470688583]]]]]<br><br><br><br><br>    [[[[[0.9493285525143782 0.852067116006533]]]<br><br><br>      [[[0.7415341862100875 0.7998965644513877]]]]<br><br><br><br>     [[[[0.92
```
mikapfl, line 562: please remove all outputs from the usage notebook.

primap2/_aggregate.py
```diff
@@ -267,51 +294,109 @@ def aggregate_coordinates(
...
         # timeseries that are aggregated
         da_out = self._da.pr.dequantify()
 
         for coordinate in agg_info.keys():
             aggregation_rules = agg_info[coordinate]
-            full_coord_name = da_out.pr.dim_alias_translations[coordinate]
+            if coordinate in da_out.pr.dim_alias_translations.keys():
+                full_coord_name = da_out.pr.dim_alias_translations[coordinate]
+            else:
+                full_coord_name = coordinate
```
mikapfl, line 307:
```suggestion
            full_coord_name = da_out.pr.dim_alias_translations.get(coordinate, coordinate)
```

primap2/_aggregate.py
```diff
@@ -267,51 +294,109 @@ def aggregate_coordinates(
...
-                        coords={
-                            full_coord_name: (full_coord_name, [value_to_aggregate])
-                        }
+                # check if all source values present
+                values_present = list(da_out[full_coord_name].values)
+                source_values_present = [
+                    val for val in source_values if val in values_present
+                ]
+                missing_values = list(set(source_values) - set(source_values_present))
+                if len(missing_values) > 0:
```
mikapfl, line 350:
```suggestion
                if missing_values:
```

primap2/_aggregate.py
```diff
@@ -267,51 +294,109 @@ def aggregate_coordinates(
...
-                    data_agg = data_agg.assign_coords(
-                        coords={
-                            full_coord_name: (full_coord_name, [value_to_aggregate])
-                        }
+                # check if all source values present
+                values_present = list(da_out[full_coord_name].values)
+                source_values_present = [
+                    val for val in source_values if val in values_present
+                ]
+                missing_values = list(set(source_values) - set(source_values_present))
```
mikapfl, line 349:
```suggestion
                missing_values = set(source_values) - set(source_values_present)
```
no need to convert back to list

primap2/_aggregate.py
```diff
@@ -267,51 +294,109 @@ def aggregate_coordinates(
...
+                        f"Not all source values present for "
+                        f"{value_to_aggregate!r} in coordinate {full_coord_name!r}. "
+                        f"Missing: {missing_values}"
+                    )
+                if source_values_present:
+                    filter.update({coordinate: source_values_present})
+                    data_agg = da_out.pr.loc[filter].pr.sum(
+                        dim=coordinate, skipna=skipna, min_count=min_count
+                    )
+                    if not data_agg.isnull().all().data:
```
mikapfl, line 361: What does `.data` do in this context? It is used to access the underlying datastructure for xarray, which in our context usually is a numpy array, but could also be something else, like a dask array or a netcdf file backed array or whatever. Maybe you don't need it at all, or just need to call `item()` to get a simple bool?

primap2/_aggregate.py
```diff
@@ -717,112 +802,209 @@ def fill_na_gas_basket_from_contents(
             )
         )
 
-    def aggregate_coordinates(
+    def add_aggregates_coordinates(
```
mikapfl, line 805: Do you mean `add_aggregated_coordinates` (with a **d** instead of an **s**)? It sounds a bit weird as is.

primap2/_data_format.py
```diff
@@ -289,9 +289,15 @@ def add_dim(
         terminology: str | None = None,
     ) -> xr.Dataset:
         """
+        TODO: should we rename this to expand_dims to indicate that it does the
+         same as the xarray function, just for primap2 format?
+
         Add a dimension and coordinate to dataset and if a terminology is given also to attrs.
```
mikapfl, line 295: definitely. That will be less confusion. If we need legacy compatibility, we can keep the `add_dim` as an alias

mikapfl (approved):
Nice, I think this can be the last round of changes.

primap2/_aggregate.py
JGuetschow: It works without `.data` Maybe it was leftover from code restructuring

primap2/_aggregate.py
JGuetschow: It's named `add_aggregates_coordinates` on purpose as I add aggregates on coordinated and don't aggregate coordinates (which in my mind would be some sort of dimensionality reduction). but I agree the name sounds a bit weird.

mikapfl: Ahh, "aggregates" is a noun. Yeah, that makes sense.

JGuetschow:
Something in the suggested changes broke the code. I'll find what cased the issue and will push a fix

primap2/_data_format.py
JGuetschow: I've done the renaming. An alias is not necessary the legacy of the function is currently very limited, so I can change the calls

PR #231, opened 2024-05-02
Properly use variables and entities in definitions and skipping

mikapfl:
# Pull request

- [x] Tests added

## Description

* `entity` and `variable` are defined things in primap2, actually use the terms correctly in the code and interfaces.
* Allow filtering by `entity` and `variable` in strategy and priority definitions.
* Better error messages if undefined dimensions are used in strategy definitions.


JGuetschow (approved):
looks all good. I've added a test because when I used my example it looked as if there was a problem (was not in the primap2 code) and used `equal_or_in` in one additional place

PR #236, opened 2024-06-03
Fix pr.set() existing=overwrite with NaN-containing value

mikapfl:
# Pull request

- [x] Tests added

## Description

Previously, `da.pr.set(dim, key, value, existing="overwrite")` would not correctly overwrite data points from da if the corresponding data point in `value` was NaN. Now it does.


mikapfl:
Doesn't solve the original problem you had, @JGuetschow , but at least fixes the bug with NaN.

.github/workflows/ci.yml
```diff
@@ -40,8 +40,7 @@ jobs:
     - name: Test with pytest (windows)
       if: startsWith(matrix.os, 'windows')
       run: |
-        .venv\Scripts\activate
-        pytest --xdoc --cov=primap2 --cov-report=xml
+        pytest --cov=primap2 --cov-report=xml
```
mikapfl, line 43: This is wrong, apparently I misunderstood the merging interface of github, will correct

mikapfl:
Sorry for the noise, now it should be ready for review @crdanielbusch .

mikapfl:
The diff is a bit hard to read because by changing the way the overwriting is handled, I broke several doctests, which I then added as real tests and fixed by further changes.

PR #241, opened 2024-06-17
workaround for xarray additional coordinate bug

JGuetschow:
This pr adds a workaround for the issue with additional (non-indexed) coordinates in some xarray functions (https://github.com/pydata/xarray/issues/9124). 

There are also some small improvements of error messages included. 

# Pull request

- [x] Tests added
- [ ] Documentation added (where applicable) <- not applicable here

## Description
The workaround is to add custom `pr.fillna` and `pr.combine_first` methods which wrap around `fillna` and `combine_first` and take care of the additional coordinate. 

The improvements to error messages are to give more relevant information on the data that caused the error.

### DIFF
--- a/primap2/_fill_combine.py
+++ b/primap2/_fill_combine.py
@@ -1,0 +1,207 @@
+"""workaround for xarray coordinate bug
+
+Additional coordinates are lost during operations that require aligning if the
+coordinates differ for the structures to be aligned.
+
+see https://github.com/pydata/xarray/issues/9124
+
+"""
+
+import xarray as xr
+from loguru import logger
+
+from ._accessor_base import BaseDataArrayAccessor, BaseDatasetAccessor
+
+
+class DataArrayFillAccessor(BaseDataArrayAccessor):
+    def fillna(self: xr.DataArray, da_fill: xr.DataArray) -> xr.DataArray:
+        """
+        Wrapper for `fillna` which ensures that additional coordinates present in
+        the calling DataArray are also present in the result. The default xarray
+        `fillna` implementation silently drops additional (non-indexed) coordinates
+        if they need alignment (do not cover the same values in both DataArrays)
+
+
+        Parameters
+        ----------
+        da_fill
+            DataArray used to fill nan values in the calling DataArray
+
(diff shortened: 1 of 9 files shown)


primap2/_fill_combine.py
```diff
@@ -0,0 +1,207 @@
...
+            except Exception as ex:
+                message = f"Could not re-add lost coordinate {coord}: {ex}"
+                logger.error(message)
+                raise ValueError(message) from ex
+
+        return filled
+
+
+class DatasetFillAccessor(BaseDatasetAccessor):
+    # fillna, combine_first
```
mikapfl, line 104:
```suggestion
```

primap2/_merge.py
```diff
@@ -88,6 +89,13 @@ def generate_log_message(da_error: xr.DataArray, tolerance: float) -> str:
     * convert the rest into a pandas dataframe for nice printing
     """
     scalar_dims = [dim for dim in da_error.dims if len(da_error[dim]) == 1]
+    scalar_dims_format = []
+    for dim in scalar_dims:
+        if dim == "time":
+            single_date = date.fromtimestamp(da_error[dim].item() / 1000000000)
+            scalar_dims_format.append(f"{dim}={single_date.strftime('%Y')}")
```
mikapfl, line 96: What's happening here?

mikapfl (approved):

primap2/_fill_combine.py
JGuetschow: ups, forgot to delete my todo list

primap2/_merge.py
JGuetschow: converting the timecode to something human readable (a year). It looks nasty so probably I missed the correct function to use.

mikapfl: For now, maybe just add a comment which explains what's going on. I also suspect that it doesn't work if you use another resolution for your datetime, but I don't think it's super important.

JGuetschow: But there was also a bug and the formatting was never used. Fixed now

Issue #255, opened 2024-08-28
Python 3.12 compatibility


mikapfl:
We haven't updated primap2 so that it works with python 3.12. We should fix that.

Issue #205, opened 2024-04-02
Unit handling in xarray fails for Python 3.12


crdanielbusch:
**Describe the bug**

The function [`pr.sum`](https://github.com/pik-primap/primap2/blob/95d96f3d1969ebd8feca4c00774006aacde68c03/primap2/_aggregate.py#L111) removes the units in a dataset for Python version 3.12.2

**Failing Test**

[github_issue_ds_1.csv](https://github.com/pik-primap/primap2/files/14835089/github_issue_ds_1.csv)

```Python
import primap2 as pm

test_data = pm.open_dataset("github_issue_ds_1.csv")

test_data = test_data.pr.sum(dim="category", skipna=True, min_count=1)

print(test_data.pr.to_interchange_format()['unit'])
```

Output for Python 3.10.13:

```
0      CH4 * gigagram / a
1       CO * gigagram / a
2      CO2 * gigagram / a
3      N2O * gigagram / a
4    NMVOC * gigagram / a
5      NOx * gigagram / a
6      SO2 * gigagram / a
Name: unit, dtype: object
```

Output for Python 3.12.2:

```
/Users/danielbusch/Documents/UNFCCC_non-AnnexI_data/venv/lib/python3.12/site-packages/xarray/namedarray/core.py:215: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  return NamedArray(dims, np.asarray(data), attrs)
/Users/danielbusch/Documents/UNFCCC_non-AnnexI_data/venv/lib/python3.12/site-packages/xarray/namedarray/core.py:215: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  return NamedArray(dims, np.asarray(data), attrs)
/Users/danielbusch/Documents/UNFCCC_non-AnnexI_data/venv/lib/python3.12/site-packages/xarray/namedarray/core.py:215: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  return NamedArray(dims, np.asarray(data), attrs)
/Users/danielbusch/Documents/UNFCCC_non-AnnexI_data/venv/lib/python3.12/site-packages/xarray/namedarray/core.py:215: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  return NamedArray(dims, np.asarray(data), attrs)
[...]

PR #259, opened 2024-09-12
maintenance: add python 3.12 compatibility, uv, dependencies

mikapfl:
# Pull request

- [x] Tests added

## Description

* Add python 3.12 compatibility - closes: #205 closes: #255
   This only needed updates in our dependencies, so easy to do.
* Tighten dependencies - we now have minimum supported versions for all dependencies.
* Add testing with the lowest supported versions to github action workflows
* Use `uv pip` instead of `pip` everywhere, it is faster
* optimize github actions for more caching


mikapfl:
This was surprisingly easy to do, until I started testing on Windows, then it was surprisingly hard to do.

mikapfl:
We now do caching via `uv` and it works quite well. This is also more important now that we have a larger test matrix, which also involves somewhat unusual combinations of older library versions on newer python versions. These need to be compiled often because binary wheel are not available, which takes quite long, but is nicely cached by `uv`.

mikapfl:
I also re-enabled xdoctests, they were relatively easy to fix and are useful after all.

crdanielbusch (approved):

PR #260, opened 2024-09-12
fix: solve warnings

mikapfl:
# Pull request

- [ ] Tests added

## Description

Solve all warnings in a current `make test` run.

* [x] `ds.dims` warning
* [x] Nanosecond timestamp resolution warning
* [x] globalwarmingpotential deprecated importlib

### DIFF
--- a/primap2/tests/test_data_format.py
+++ b/primap2/tests/test_data_format.py
@@ -45,7 +45,7 @@ class TestEnsureValid:
     def test_time_dimension_for_metadata(self, opulent_processing_ds, caplog):
         opulent_processing_ds["Processing of CO2"] = opulent_processing_ds[
             "Processing of CO2"
-        ].expand_dims(dim={"time": np.array(["2020", "2021"], dtype=np.datetime64)})
+        ].expand_dims(dim={"time": np.array(["2020", "2021"], dtype="datetime64[ns]")})
         with pytest.raises(ValueError, match=r"contains metadata, but carries 'time' dimension"):
             opulent_processing_ds.pr.ensure_valid()
         assert "ERROR" in caplog.text
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,3 +25,7 @@ known-first-party = ["primap2"]
 convention = "numpy"
 
 [tool.setuptools_scm]
+
+[tool.pytest.ini_options]
+# actually, importlib.resources.open_text will be undeprecated again in python 3.13
+filterwarnings = "ignore:.*open_text is deprecated.*:DeprecationWarning"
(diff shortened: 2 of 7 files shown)


mikapfl:
The idea is to first solve all warnings in a `make test`, then go into the whole pandas and numpy situation.

JGuetschow:
Sounds good. `ds.dims` is also one of the most dangerous looking of warnings that I get often (I think it will actually break things).

mikapfl:
Yeah, ds.dims will break things (and the transition is a bit annoying, I think they should've instead left `ds.dims` as inconsistent with `da.dims` and invent something new like `ds.dim` and `da.dim` or so for the consistent thing).
So I guess we should also export the dim_names function stand-alone so that it can be used from other code? Do you need this in code outside primap2?

We could also add this to our accessors as `ds.pr.dim_names` and `da.pr.dim_names`.

JGuetschow:
I current mostly use the coordinates  (`ds.coords`) from the outside, not the dimension names

mikapfl:
Also, when they turn ds.dims into an iterable of hashables ("list of str", to its friends), `ds.dims.keys()` (what we were using mostly) will just throw an exception, so at least no wrong results.

primap2/tests/test_data_format.py
```diff
@@ -45,7 +45,7 @@ def test_valid_ds_pass(self, any_ds, caplog):
     def test_time_dimension_for_metadata(self, opulent_processing_ds, caplog):
         opulent_processing_ds["Processing of CO2"] = opulent_processing_ds[
             "Processing of CO2"
-        ].expand_dims(dim={"time": np.array(["2020", "2021"], dtype=np.datetime64)})
+        ].expand_dims(dim={"time": np.array(["2020", "2021"], dtype="datetime64[ns]")})
```
JGuetschow, line 48: What is the reason to pass dtype as a string?

pyproject.toml
```diff
@@ -25,3 +25,7 @@ known-first-party = ["primap2"]
 convention = "numpy"
 
 [tool.setuptools_scm]
+
+[tool.pytest.ini_options]
+# actually, importlib.resources.open_text will be undeprecated again in python 3.13
```
JGuetschow, line 30: Crazy

JGuetschow (approved):
Looks good to me. Just a question out of curiosity regarding passing the dtype a string

primap2/tests/test_data_format.py
mikapfl: if you want to specify nanosecond resolution for the dtype, it is not possible to write `np.datetime64[np.ns]` or something like that, I think. Anyway, it was easier to figure out how to specify the nanosecond resolution for the dtype as a string.

pyproject.toml
mikapfl: Yes, it was actually scheduled for removal in python 3.13, but at the last minute someone contributed a solution which made it possible to maintain the function going forward instead.

crdanielbusch (approved):

Issue #256, opened 2024-08-28
get ready for upcoming pandas 3.0 changes


mikapfl:
We should support pandas >= 2.0, and pandas >= 2.0 only. There should be no deprecation warnings when running our tests.

JGuetschow:
I assume you mean support >= 3.0?

mikapfl:
Am I confused? According to https://pandas.pydata.org/docs/whatsnew/index.html , the latest pandas version is 2.2.

mikapfl:
Ah, are you talking about https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#upcoming-changes-in-pandas-3-0 ?

That's probably also worth exploring, to make primap2 ready for pandas 3 as much as possible.

JGuetschow:
Yes, that's what's currently causing all (or most of) the warnings and deprecation notices.

PR #261, opened 2024-09-13
Use pandas 2

mikapfl:
# Pull request

- [ ] Tests added

## Description

Updated our pandas dependency to version 2

closes: #256


mikapfl:
Only very minimal fixes were necessary to have passing and warnings-free tests locally. Now investigating what happens on windows CI.

mikapfl:
Yeah, we're not yet ready for numpy 2. For downscaling we need certain accelerated functions in xarray, which can be obtained from either the bottleneck package or the numbagg package. But neither bottleneck nor numba are ready for numpy 2, yet. So we will have to wait a bit more for numpy 2.

JGuetschow:
The warning come from the `pm2io` module. But I think from the tested functions like `from_interchange_format`. I'll check again.

JGuetschow:
```
primap2/pm2io/_GHG_inventory_reading.py:166: FutureWarning: The previous implementation of stack is deprecated and will be removed in a future version of pandas. See the What's New notes for pandas 2.1.0 for details. Specify future_stack=True to adopt the new implementation and silence this warning.
  df_stacked = df_nir.stack([0, 1], dropna=True).to_frame()
```
So far I could always just adopt the new implementation without problems, but I've not checked yet if that works in all cases. I assume not because the warning would be useless if there's nothing to warn about

I also get warnings about setting values in chained assignments, but I don't have an example for a primap2 function throwing them. If I find one I'll add it here.

mikapfl:
We discussed that we don't need to fix the warnings in `_GHG_inventory_reading.py` because the function is only needed for reading deprecated report formats.

PR #269, opened 2024-10-10
Add category conversion

crdanielbusch:
# Pull request

- [ ] Tests added

## Description

Continues Mika's work from https://github.com/primap-community/primap2/pull/74. We want to be able to convert categories from one standard into another, for example from IPCC1996 to IPCC2006. More info here https://github.com/primap-community/climate_categories/issues/160


primap2/tests/data/BURDI_conversion.csv
```diff
@@ -0,0 +1,40 @@
+# references: non_annex1_data repo
+# last_update: 2024-10-14
```
crdanielbusch, line 2: @JGuetschow does the `BURDI` to `IPCCC2006_PRIMAP` conversion make sense to you? I took took the rules from the [config in the UNFCCC_non-AnnexI_data repo](https://github.com/JGuetschow/UNFCCC_non-AnnexI_data/blob/main/src/unfccc_ghg_data/unfccc_di_reader/unfccc_di_reader_config.py). The rules here combine the mapping and aggregation step. However we still need some aggregation after the conversion, for example for `"3": {"sources": ["M.AG", "M.LULUCF"]`. That's beyond the scope of what the conversion should do if I remember correctly?

primap2/_convert.py
```diff
@@ -12,12 +12,13 @@
...
-    @alias_dims(["dim"])
-    def convert(
+    def _convert_inner(
         self,
         dim: Hashable | str,
-        categorization: climate_categories.Categorization | str,
         *,
+        # TODO type will change to climate_categories.Conversion when
+        #  https://github.com/primap-community/climate_categories/pull/164 is merged
+        conversion: climate_categories._conversions.Conversion,
```
mikapfl, line 21: Now that I see the definition of `_convert_inner`, I think we can go all-in with this approach. As you said in chat, we can leave the whole "get the right conversion" business to `climate_categories`, where we have functions for this already. So, we just call this here `convert` and throw away the wrapper which fetches the conversion from a name. The resulting call for a "simple" conversion would look like this:

```python
import climate_categories as cc

conversion = cc.BURDI.conversion_to(cc.IPCC2006_PRIMAP)
da.convert("category", conversion=conversion)
```

for me, this is actually more transparent than what the current `convert` does:

```python
da.convert("category", "IPCC2006_PRIMAP")
```

Sure, the current `convert` function needs less typing, but it also hides where the conversions come from, so as soon as you define your own categorization which is not built into climate_categories, it is unclear how you would use it. In contrast, using the "inner" function as the public API, it is obvious that you can build your own conversion and pass that.

mikapfl: I think I originally designed the API of `convert` to be similar to e.g. GWP and unit conversion. However, for units and also global warming potentials, we assume that user-defined units and GWPs will be super uncommon, so hiding the complexities from the user made sense. So, users don't have to pass a `Unit` object, they can simply pass a string with a name.

However, here, I think we already know from the non-AnnexI repo that we will frequently have to modify the `conversion` before we can use (or use a fully custom conversion). In this case, users have a `Conversion` object already anyway, so it makes sense to use the `Conversion` object instead of names.

But we should check that the passed `Conversion` actually converts from the actual source categorization of the data and refuse to convert FAOstat categories using a conversion from BURDI to IPCC2006.

JGuetschow: makes all sense to me

primap2/tests/data/BURDI_conversion.csv
```diff
@@ -0,0 +1,40 @@
...
+2.B + 2.E,2.B
+2.C,2.C
+2.F,2.F
+2.G + 2.D, 2.H
+3,2.D
+4,M.AG
+4.A,3.A.1
+4.B,3.A.2
+4.C,3.C.7
+4.D + 4.C + 4.E + 4.F + 4.G,3.C
```
JGuetschow, line 25: It would be good to also map 4.D to single category, else we have some of the 3.C subcategories but not all which can lead to confusion and problems in category aggregation and consistency checks. I would add an M-category to IPCC2006_PRIMAP

primap2/tests/data/BURDI_conversion.csv
JGuetschow: You could add `4 + 5, 3` as a rule to aggregate AFOLU as you already do the same for  2 and 3. I think we just wanted to exclude downscaling, not aggregation.

JGuetschow (commented):
I've added a few comments. Have to submit a review so they are actually visible.

primap2/_convert.py
crdanielbusch: Thanks for the input Mika and Johannes. Mika if you could have a quick look at `_convert.py`, that would be great. I have now integrated the changes from the latest climate categories release. In the end, only a few lines of code changed.

mikapfl:
Before you merge, you need to change the target of the PR to `main` (currently `conversions`).

primap2/_convert.py
```diff
@@ -34,10 +34,10 @@ def convert(
...
         dim : str
             Dimension to convert. Has to be a dimension from ``da.dims``.
-        categorization : climate_categories.Categorization or str
-            New categorization to convert the given dimension to. Either give the title
-            of the new categorization (like ``IPCC1996``) or a
-            ``climate_categories.Categorization`` object.
+        conversion : climate_categories.Conversion
+            The conversion rules that describe the conversion from the old to the new
+            categorization. Contains ``climate_categories.Categorization``
+            object for old and new categorization.
```
mikapfl, line 40:
```suggestion
            categorization. Contains the ``climate_categories.Categorization``
            objects for the old and the new categorization.
```

mikapfl (approved):
Nice, definitely a cleaner API now, and adding the tests is super useful for further development. :slightly_smiling_face:

mikapfl:
Changes we also discussed which aren't in this PR yet, but will be much easier now with the added tests:
* remove the `sum_rule` argument and code, always assume extensive. Add documentation that this is only meant to be used with extensive quantities.
* remove the `input_weights` and `output_weights` arguments and code. Since we always assume extensive, the `input_weights` are never needed, and we decided we don't want to do downscaling in the category conversion function, so `output_weights` are also not needed. We have to deal with multiple categories in the target differently. Ideas we had so far:
  - e.g. if you have categories `1.A` and `1.B` on the rhs, automatically add an `A.1.A+1.B` category (name up to debate).
  - if there is a category which if the sum of the categories on the rhs already, and the rhs categorization is `total_sum`, simply use the existing super-category. E.g. if the rule specifies categories `1.A` and `1.B`, and in the rhs categorization `1` has the children `1.A` and `1.B` only and is `total_sum`, simply use `1` as the rhs category.
  - maybe don't add automatic `A.` categories in the conversion, raise an error instead and the user has to add appropriate `M.` categories to the categorization before the conversion.

Probably, these changes deserve their own PR each. Also, still some open questions around best usability re the downscaling question. My suggestion: In a first PR remove the `sum_rule`, in a second PR which builds on the first remove the `input_weights` and `output_weights` and raise errors instead if downscaling would be necessary. Then, start work on an actual conversion and add downscaling support as needed. Maybe we don't need anything automatic, maybe some automation is useful, but that will be much clearer once e.g. the FAO conversion is written and adding also these super-categories by hand is annoying or actually rather pleasant.

PR #272, opened 2024-10-21
Harmonize filtering / selection parameter names

JGuetschow:
…other functions

# Pull request

- [x] Tests added

## Description

### DIFF
--- a/primap2/_aggregate.py
+++ b/primap2/_aggregate.py
@@ -310,22 +310,22 @@ class DataArrayAggregationAccessor(BaseDataArrayAccessor):
                     else:
                         rule_tolerance = tolerance
                     if "sel" in rule.keys():
-                        sel_ = rule.pop("sel")
-                        if "variable" in sel_.keys():
-                            if da_out.name in sel_["variable"]:
-                                sel_.pop("variable")
+                        sel = rule.pop("sel")
+                        if "variable" in sel.keys():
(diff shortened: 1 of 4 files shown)


primap2/_aggregate.py
```diff
@@ -309,23 +309,23 @@ def add_aggregates_coordinates(
                         rule_tolerance = rule.pop("tolerance")
                     else:
                         rule_tolerance = tolerance
-                    if "filter" in rule.keys():
-                        filter_ = rule.pop("filter")
-                        if "variable" in filter_.keys():
-                            if da_out.name in filter_["variable"]:
-                                filter_.pop("variable")
+                    if "sel" in rule.keys():
+                        sel_ = rule.pop("sel")
```
mikapfl, line 313:
```suggestion
                        sel = rule.pop("sel")
```
`filter` is a python reserved keyword (well, just a built-in function name, to be precise), that's the only reason we wrote `filter_` with the `_` at the end. Since `sel` is not a built-in function name, we can simply call it `sel`.

mikapfl (approved):
Looks good, thanks. Just rename `sel_` to `sel`.

primap2/_aggregate.py
JGuetschow: Will do, was lazy. the variable was called `filter_` and I just replaced `filter` with `sel`

PR #275, opened 2024-10-21
Test development versions of upstream libraries

mikapfl:
# Pull request

- [ ] Tests added

## Description

* add a dependency group which installs development versions of central upstream libraries
* add a github actions workflow which tests using the development dependency group
* update versions of gha actions, also upload test results to codecov

### DIFF
--- a/.github/workflows/ci-upstream-dev.yml
+++ b/.github/workflows/ci-upstream-dev.yml
@@ -1,10 +1,12 @@
 name: CI upstream development versions
 
 on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch: # allows you to trigger the workflow run manually
   schedule:
     - cron: "5 2 * * tue" # Tuesdays at 02:05 UTC.
-  workflow_dispatch: # allows you to trigger the workflow run manually
-  pull_request:
 
 jobs:
   test:
--- a/.github/workflows/ci-upstream-dev.yml
+++ b/.github/workflows/ci-upstream-dev.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: "5 2 * * tue" # Tuesdays at 02:05 UTC.
   workflow_dispatch: # allows you to trigger the workflow run manually
+  pull_request:
 
 jobs:
   test:
--- a/.github/workflows/ci-upstream-dev.yml
+++ b/.github/workflows/ci-upstream-dev.yml
@@ -34,6 +34,14 @@ jobs:
       run: |
         uv pip install --resolution highest --prerelease allow .[test,upstream-dev]
 
-    - name: Test with pytest
+    - name: Test with pytest (linux)
+      if: startsWith(matrix.os, 'ubuntu')
       run: |
-        uv run pytest --xdoc
+        source .venv/bin/activate
+        pytest --xdoc
+
+    - name: Test with pytest (windows)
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-upstream-dev.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Install ${{ matrix.resolution }} dependencies
+      run: |
+        uv pip install --resolution highest --prerelease allow .[test,upstream-dev]
+
+    - name: Test with pytest (linux)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        source .venv/bin/activate
+        pytest --xdoc
```
znicholls, line 41:
```suggestion
        uv run pytest --xdoc
```

Or does this not work for a reason I can't see?

.github/workflows/ci-upstream-dev.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Test with pytest (windows)
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        .venv\Scripts\activate
+        pytest --xdoc
```
znicholls, line 47:
```suggestion
        uv run pytest --xdoc
```

(which then means you don't need the windows specific step) Or does this not work for a reason I can't see?

.github/workflows/ci.yml
```diff
@@ -14,7 +15,7 @@ jobs:
         python-version: [ "3.10", "3.11", "3.12" ]
         resolution: [ "highest", "lowest-direct" ]
```
znicholls, line 16: Clever

setup.cfg
```diff
@@ -59,6 +59,15 @@ test =
...
     xdoctest>=1.2
+# install development versions of central upstream libraries
+upstream-dev =
+    pint@git+https://github.com/hgrecco/pint
+    pint_xarray@git+https://github.com/xarray-contrib/pint-xarray
+    xarray@git+https://github.com/pydata/xarray
+    pandas@git+https://github.com/pandas-dev/pandas
+    openscm_units@git+https://github.com/openscm/openscm-units
+    # numpy dev currently not supported
+    #numpy@git+https://github.com/numpy/numpy
```
znicholls, line 70: Nice. Seems the cleanest way to do this I think.

znicholls (approved):
Very nice. I'll be interested to see how this plays out because it would be a super nice pattern to reproduce across multiple repos.

.github/workflows/ci-upstream-dev.yml
mikapfl: I think uv is installed in the venv, so things are somewhat circular. But maybe not, I'll test!

mikapfl: No, it fails because we haven't fully adopted uv yet, just surgically replaced `pip` with `uv pip` and `venv` with `uv venv`. I tried fully adopting uv when I did that, but building was hard or something, after messing around for half an hour I decided to postpone that for a few more months until uv can do everything.

See the failing CI here: https://github.com/primap-community/primap2/actions/runs/11445079855/job/31841383315?pr=275

znicholls: Ah ok fair, makes sense. Thanks!

crdanielbusch (approved):

PR #281, opened 2024-10-28
remove sum rule from conversion

crdanielbusch:
# Pull request

- [x] Tests added

## Description

Previously the `convert` function supported extensive data (like total emissions in a year subdivided into multiple
sectoral categories) and intensive data (like average per-person emissions in a year subdivided into different territorial entities). We decided that we don't want to support the handling of intensive data, as we don't 
expect this to be a common use case.

This pull request removes the `sum_rule` argument and code and always assumes extensive data. It adds documentation that this is only meant to be used with extensive quantities.

The general plan (Mika): In a first PR remove the `sum_rule`, in a second PR which builds on the first remove the `input_weights` and `output_weights` and raise errors instead if downscaling would be necessary. Then, start work on an actual conversion and add downscaling support as needed.

### DIFF
--- a/primap2/_convert.py
+++ b/primap2/_convert.py
@@ -389,7 +389,7 @@ def factors_categories_to_xarray(
     """Convert dictionary mapping categories to factors into xarray-compatible objects.
 
     Using the xarray objects ensures that in subsequent calculations, everything
-    will cleanly multiply reagardless of the dimensionality of the data.
+    will cleanly multiply regardless of the dimensionality of the data.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


primap2/_convert.py
```diff
@@ -472,7 +449,6 @@ def derive_weights(
     dim: str,
```
crdanielbusch, line 449: `derive_weights` function will change again when I remove `input_weights` and `output_weights`

crdanielbusch:
@JGuetschow do you wanna have a look at this? The changes are relatively simple and don't really change the conversion algorithm. The more interesting part will be in the next PR when we deal with multiple categories in the target categorisation.

JGuetschow (approved):
Looks all good to me. 
Was 'extensive' the only implemented rule? I would expect removed code and tests for other rules.

crdanielbusch:
Thanks for the review.

> I would expect removed code and tests for other rules.

There were two different options, `extensive` for total ghg-emissions, and `intensive` for ghg-emission concentrations (per GDP, per capita etc.). We knew that we don't really need the `intensive` case, so I wrote all the test for the `extensive` case.

Without the `intensive` option, the logic in the `derive_weights` function is simplified: the 1-to-1 case is trivial, the n-to-1 case means we sum up, and the 1-to-n case means we need weights on the target side. The weights for the target side will be removed in the next PR, so I didn't spend to much time on developing this part of the code.

JGuetschow:
OK, so there just were no tests for the `intensive` case and thus none had to be removed. thanks.

PR #282, opened 2024-10-28
remove input and output weights from convert function

crdanielbusch:
# Pull request

Please confirm that this pull request has done the following:

- [x] Tests added
- [ ] Documentation added (where applicable)
- [ ] Description in a `{pr}.thing.md` file in the directory `changelog` added - see [changelog/README.md](https://github.com/pik-primap/primap2/blob/main/changelog/README.md) for details

## Description

This pull request builds on #281 which should be merged first. Removes the `input_weights` and `output_weights` from the convert function and raise errors if downscaling would be necessary. 

At this point only one-to-one, and many-to-one conversions are possible.

In #284 support for one-to-many conversions is added again.

PR #284, opened 2024-10-29
one-to-many conversion

crdanielbusch:
# Pull request

- [x] Tests added

## Description

Functionality added in this pull request:
- **one-to-many conversion**: Input and output weights are not supported anymore (see PR #282). That means a category (or a group of categories) cannot be split into a group of categories, e.g. `4.B.10 + 4.B.11 + 4.B.12 + 4.B.13 -> 3.A.2.j` works, but `2.C.5 -> 2.C.5 + 2.C.6 + 2.C.7` does not work. When there are several categories on the target side, we now create a new category, e.g. the rule `2.C.5 -> 2.C.5 + 2.C.6 + 2.C.7` would yield a new category `M_2.C.5_2.C.6_2.C.7`. 
- **one-to-one conversion but target category is not part of target categorisation**: If a category on the target side does not exist it we need to create it in the respective categorisation in climate categories. Sometimes categories don't have a matching target category or group of categories, but the information should not be lost in the target categorisation, e.g. `4.D -> M.3.C.45.AG` where `M.3.C.45.AG` is the sum of the agriculture-related emissions of `3.C.4` and `3.C.5`. In this case the user needs to add the category manually to the categorisation in climate categories. We think this will be needed only for `IPCC2006_PRIMAP`. 

**Questions:**
- Does it make sense that a one-to-many conversion creates a new category automatically but an unknown single category needs to be added to climate categories? For example, you could write in your conversion `2.C.5 -> 2.C.5 + 2.C.6 + 2.C.7` and the new category `M_2.C.5_2.C.6_2.C.7` will be created (if all categories are part of your target categorisation). At the same time the `4.D -> M.3.C.45.AG` will raise an error, because `M.3.C.45.AG` is not part of the target categorisation.
- At what point should we move into a real world example (FAO) instead of thinking about theoretical examples. Personally, I would rather have a simple conversion function and then develop it further as we read the FAO data.

**Background (Mika):**
- We have to deal with multiple categories in the target differently. Ideas we had so far:
    - e.g. if you have categories `1.A` and `1.B` on the rhs, automatically add an `A.1.A+1.B` category (name up to debate). 
    - if there is a category which if the sum of the categories on the rhs already, and the rhs categorization is `total_sum`, simply use the existing super-category. E.g. if the rule specifies categories `1.A` and `1.B`, and in the rhs categorization `1` has the children `1.A` and `1.B` only and is `total_sum`, simply use `1` as the rhs category.
    - maybe don't add automatic `A.` categories in the conversion, raise an error instead and the user has to add appropriate `M.` categories to the categorization before the conversion.

- In this PR: start work on an actual conversion and add downscaling support as needed. Maybe we don't need anything automatic, maybe some automation is useful, but that will be much clearer once e.g. the FAO conversion is written and adding also these super-categories by hand is annoying or actually rather pleasant.

### DIFF
--- a/primap2/_convert.py
+++ b/primap2/_convert.py
@@ -186,7 +186,8 @@ class DataArrayConversionAccessor(_accessor_base.BaseDataArrayAccessor):
             # if there is more than one category on the target side
             if len(output_selection[new_dim]) > 1:
                 # TODO this leads to very long category names
-                new_category = "M_" + "_".join(output_selection[new_dim])
+                new_category = create_category_name(rule)
+                # new_category = "A_(" + "_".join(output_selection[new_dim]) + ")"
                 # add newly created category to da
                 new_categories = list(da.indexes["category (IPCC2006)"]) + [new_category]
                 da = da.reindex({"category (IPCC2006)": new_categories}, fill_value=np.nan)
@@ -432,3 +433,12 @@ def prepare_auxiliary_dimensions(
     return {
         climate_categories.cats[name]: auxiliary_dimensions[name] for name in auxiliary_dimensions
     }
+
+
+def create_category_name(rule):
+    factor_to_string = {1: "+", -1: "-"}
+    components = [factor_to_string[i[1]] + i[0].codes[0] for i in rule.factors_categories_b.items()]
+    # remove the first "+" sign in the name (leave a "-" sign in)
+    if components[0][0] == "+":
+        components[0] = components[0][1:]
+    return "A_(" + "".join(components) + ")"
--- a/primap2/_convert.py
+++ b/primap2/_convert.py
@@ -185,10 +185,7 @@ class DataArrayConversionAccessor(_accessor_base.BaseDataArrayAccessor):
 
             # if there is more than one category on the target side
             if len(output_selection[new_dim]) > 1:
-                # TODO this leads to very long category names
                 new_category = create_category_name(rule)
-                # new_category = "A_(" + "_".join(output_selection[new_dim]) + ")"
-                # add newly created category to da
                 new_categories = list(da.indexes["category (IPCC2006)"]) + [new_category]
                 da = da.reindex({"category (IPCC2006)": new_categories}, fill_value=np.nan)
                 new_output_selection = output_selection.copy()
@@ -435,10 +432,22 @@ def prepare_auxiliary_dimensions(
     }
 
 
-def create_category_name(rule):
+def create_category_name(rule: climate_categories.ConversionRule):
+    """
+    Create a category name based on the provided rule.
+
+    Parameters
+    ----------
+    rule : climate_categories.ConversionRule
+        rule to convert between categories from two different categorizations.
+
+    Returns
+    -------
+        The generated category name.
+    """
     factor_to_string = {1: "+", -1: "-"}
     components = [factor_to_string[i[1]] + i[0].codes[0] for i in rule.factors_categories_b.items()]
-    # remove the first "+" sign in the name (leave a "-" sign in)
+    # remove the first "+" sign in the name (leave "-" sign in)
     if components[0][0] == "+":
         components[0] = components[0][1:]
     return "A_(" + "".join(components) + ")"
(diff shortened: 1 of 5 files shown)


primap2/tests/data/BURDI_conversion.csv
```diff
@@ -17,16 +17,19 @@ BURDI,IPCC2006_PRIMAP,comment
 2.C,2.C
 2.F,2.F
 2.G + 2.D, 2.H
+2.G, 2.H.3
 3,2.D
 4,M.AG
 4.A,3.A.1
 4.B,3.A.2
 4.C,3.C.7
+4.D, 3.C.45.AG
```
crdanielbusch, line 26: This will not work unless `3.C.45.AG` is added to the IPCC2006_PRIMAP conversion in climate categories.

primap2/tests/test_convert.py
```diff
@@ -64,12 +63,44 @@ def test_convert_ipcc(empty_ds: xr.Dataset):
...
+    # all other gases should be nan
+    all_gases_but_N2O = list(result.indexes["source (gas)"])
+    all_gases_but_N2O.remove("N2O")
+    assert np.isnan(
+        result.pr.loc[{"category": mcat, "source (gas)": all_gases_but_N2O}].values
+    ).all()
+    # rule 7 -> 5
+    assert (result.pr.loc[{"category": "5"}] == 1.0 * primap2.ureg("Gg CO2 / year")).all().item()
+    # rule 2.F.6 -> 2.E + 2.F.6 + 2.G.1 + 2.G.2 + 2.G.4,
+    # rule 2.F.6 + 3.D -> 2.E + 2.F.6 + 2.G - ignored because 2.F.G already converted
```
JGuetschow, line 92: Ignoring rules can be dangerous, because if we convert more than one country data coverage is likely different for the different countries and while one rule might apply for one country another rule might be needed for another country. I know this is not the place where the ignoring goes on, but I wanted to raise the point anyway.

primap2/tests/test_convert.py
```diff
@@ -150,12 +181,24 @@ def test_convert_BURDI(empty_ds: xr.Dataset):
     assert (
         (result.pr.loc[{"category": "3.C.7"}] == 1.0 * primap2.ureg("Gg CO2 / year")).all().item()
     )
-    # 2.E + 2.B = 2.E, 2.E should not be part of new data set
+    # rule 2.E + 2.B -> 2.B
+    # 2.B is part of PRIMAP categories, but cannot be retrieved from conversion
```
JGuetschow, line 185: You mean `2.E` here?

JGuetschow (changes requested):
Looks good, but some small comments and some questions.

primap2/tests/data/BURDI_conversion.csv
crdanielbusch: Updated to `M.3.C.45.AG`, here and in https://github.com/primap-community/climate_categories/pull/169

primap2/_convert.py
```diff
@@ -175,35 +178,28 @@ def _fill_category(
...
-            da.loc[output_selection] = rhs
-
-            if not rule.is_restricted:
-                # stop processing rules for this category
+            rhs = lhs / output_factors
+
+            # if there is more than one category on the target side
+            if len(output_selection[new_dim]) > 1:
+                # TODO this leads to very long category names
+                new_category = "M_" + "_".join(output_selection[new_dim])
```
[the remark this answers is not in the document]
crdanielbusch: Category names for multiple target categories are now created in the `create_category_name` function

primap2/tests/test_convert.py
crdanielbusch: Yep

primap2/tests/test_convert.py
JGuetschow: If 2.G has only subcategories 1-4 it looks like it can be rewritten. But I did not make the mapping, so there might be some motivation behind the rules that I don't know of.

primap2/_convert.py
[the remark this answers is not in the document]
JGuetschow: great! look very good.

JGuetschow (approved):
Looks great.

PR #286, opened 2024-11-04
Add wrapper for csg

JGuetschow:
# Pull request

This PR adds a wrapper around the `primap2.csg.compose()` function that prepares input data (e.g. remove unnecessary data) and postprocesses the output (add coordinates that were used in the process with new values).

A larger test for the compose function as well a small test for a subfunction are added as well.

Docs will be extended.

Please confirm that this pull request has done the following:

- [x] Tests added
- [x] Documentation added (where applicable)
- [x] Description in a `{pr}.thing.md` file in the directory `changelog` added - see [changelog/README.md](https://github.com/pik-primap/primap2/blob/main/changelog/README.md) for details

## Description

Please provide a short description what your pull request does.

PR #288, opened 2024-11-05
Improved CSG filling strategies

JGuetschow:
# Pull request

- [x] Tests added

## Description

Add more filling strategies for the csg 
* local trend based filling: similar to primap1 method
* local least squares filling: improvement over global least squares with configurable matching area -> postponed to a future PR

## Important
Merge #286 first -> done


docs/source/usage/csg.md
```diff
@@ -162,3 +164,62 @@ category 1 "lowpop" was preferred.
 For category 0, the initial timeseries did not contain NaNs, so no filling was needed.
 For category 1, there was information missing in the initial timeseries, so the
 lower-priority timeseries was used to fill the holes.
+
+## The `create_composite_source` wrapper function
+
+The {py:func}`primap2.csg.compose` function creates a composite time series according to
+the given priorities and strategies, but it does not take care of pre- and postprocessing
+of the data. It will carry along unnecessary data and the resulting dataset will miss the
+priority coordinates. The {py:func}`primap2.csg.create_composite_source` function takes acre
```
mikapfl, line 173:
```suggestion
priority coordinates. The {py:func}`primap2.csg.create_composite_source` function takes care
```

docs/source/usage/csg.md
```diff
@@ -162,3 +164,62 @@ category 1 "lowpop" was preferred.
...
+of these steps and prepares the input data and completes the output data to a primap2 dataset
+with all desired dimensions and metadata.
+
+The function takes the same inputs as {py:func}`primap2.csg.compose` with additional input to
+define pre- and postprocessing:
+
+* **result_prio_coords** Defines the vales for the priority coordinates in the output dataset. As the
+priority coordinates differ for all input sources there is no canonical vale
```
mikapfl, line 181:
```suggestion
priority coordinates differ for all input sources there is no canonical value
```

primap2/csg/_wrapper.py
```diff
@@ -0,0 +1,132 @@
...
+    coords: Dictionary
+        Format is 'name': {'value': value, 'terminology': terminology}
+        terminology is optional
+
+    """
+    for dim in dims.keys():
+        if "terminology" in dims[dim].keys():
+            terminology = dims[dim]["terminology"]
+        else:
+            terminology = None
```
mikapfl, line 24:
```suggestion
        terminology = dims[dim].get("terminology", None)
```

JGuetschow:
Ready for review. Docs still to come, but that should be quick as the class docstring is quite detailed.

primap2/csg/_strategies/gaps.py
```diff
@@ -419,43 +420,55 @@ def calculate_scaling_factor(
...
 
-    # TODO continue here with factor calculation and treatment of special cases
-    #   (e.g. division by 0)
+
+def fill_gap(
+    ts: xr.DataArray,
+    fill_ts: xr.DataArray,
+    gap: Gap,
+    factor: np.array,
+) -> xr.DataArray:
```
mikapfl, line 458:
```suggestion
) -> xr.DataArray:
    """Fill a gap in a timeseries using data from another timeseries and factors.
    
    Parameters
    ----------
    ts: xr.DataArray
        Timeseries with a gap to be filled.
    fill_ts: xr.DataArray
        Timeseries from which to take the data to fill in the gap in ts.
    gap: Gap
        The gap that should be filled.
    factor: np.array
        Numpy array containing two floats, the scaling factor to use at the left side of the gap
        and the scaling factor to use at the right side of the gap. The data from fill_ts will be
        multiplied with the factor before filling the gap in ts. If the factors differ, a linear
        interpolation of the scaling factor between both values will be performed.
    
    Returns
    -----------
        result
            DataArray with the same size and shape as ts, but with the given gap filled from fill_ts.
```

mikapfl (approved):
There are a few TODO comments left and minor things I found.

Comprehensive tests and well documented!

primap2/csg/_strategies/local_trends.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+
+    # TODO: add fallback options so we can calculate a trend for one to also if the
+    #  other doesn't have sufficient information. similarly if we have several gaps to fill
+    #  we can use trends where possible, but not for all gaps if data is not sufficient
+    #  for all gaps
+    # fallback_option
+
+    fit_params: FitParameters = field()
+    allow_negative: bool = field()  # not implemented yet
```
mikapfl, line 105:
```suggestion
    fit_params: FitParameters
    allow_negative: bool # not implemented yet
```

it becomes a `field()` automatically, no need to specify it.

Also, should the `allow_negative` be removed while it is not implemented yet?

JGuetschow: I think ruff complained and that's why I added the field statement. But I'll try again, maybe I missed that combination.
Regarding `allow_negative`: It's implemented, just forgot to remove the comment

JGuetschow: I get `NameError: name 'fit_params' is not defined` without the field statement

mikapfl: Ah, now I see why. you are later setting `@fit_params.default`, which only works if you `field()` it. Why is that necessary, why can't you just write

```suggestion
    fit_params: FitParameters = FitParameters(trend_length=10, min_trend_points=5, trend_length_unit="YS", fit_degree=1, fallback_degree=0)
    allow_negative: bool = False
```

and delete the `@fit_params.default` and `@allow_negative.default` functions?

JGuetschow: I don't have the exact error in mind, but I think it had something to do with making it a class or instance variable.

mikapfl:
> Commits must have verified signatures.

This happens because the pre-commit CI doesn't sign their fixes. Annoying. Once all the check are green, you can "merge without waiting for requirements (bypass rules)"

primap2/csg/_strategies/local_trends.py
JGuetschow: The ruff complaint is:
`RUF009 Do not perform function call `FitParameters` in dataclass defaults`

Was quite a bit of work/research to fix it as the solution presented in the ruff docs doesn't work either for this case.  It involved me actually looking into a book (a proper physically existing book). But in the end the `attrs` docs helped if I remember correctly. At least I learned a bit about class and instance variables and that tools to make declarations easier sometimes make them more complicated.

mikapfl: hm, okay, makes sense. Always a bit harder to read with the `field` stuff, but probably it's due to `FitParameters` being a class itself.

PR #298, opened 2024-11-29
slight refactorings, docs, typos

mikapfl:
@JGuetschow it was easier for me to add my refactorings as patches instead of doing them inline in github, so I created a PR into your branch with what I would change in your branch before merging. Changes:
* regenerated api docs (happens if you run "make docs" locally")
* added LocalTrendsStrategy to CSG usage docs
* added some docstrings, moved method docstrings from the class's docstring to the method's docstring
* unified calculate_right_boundary_trend and calculate_left_boundary_trend into one function with case distinctions
* refactored get_shifted_time_value. I first did it with python lists, which is a bit simpler, but using the underlying numpy arrays without converting is much faster, so I did that in the end.
* corrected typos wherever I found them.

### DIFF
--- a/primap2/csg/_strategies/gaps.py
+++ b/primap2/csg/_strategies/gaps.py
@@ -472,15 +472,13 @@ def get_shifted_time_value(
     Returns
     -------
         time coordinate value at desired relative position
-
     """
-    # TODO: the following is not very elegant. I struggle with tasks like getting the coordinate
-    #  value of the next item in xarray
-    mask = ts.copy()
-    mask.data = mask.data * np.nan
-    mask.pr.loc[{"time": original_value}] = 1
-    mask = mask.shift(time=shift, fill_value=np.nan)
-    return mask.coords["time"].where(mask == 1, drop=True).data[0]
+    # For actually getting the index of a value in an index, it is easiest to work with
+    # the underlying numpy arrays.
+    time_points = ts["time"].values
+    original_index = np.where(time_points == original_value)[0][0]
+    new_index = original_index + shift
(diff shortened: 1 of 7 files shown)


JGuetschow:
Thanks for adding docs. I should have said that I intentionally left them for after your review so I don't have to rewrite them in case of changes. Sorry, didn't intend to make you write the docs.

primap2/csg/_strategies/gaps.py
```diff
@@ -12,23 +14,16 @@ class Gap:
 
     Attributes
     ----------
-    type :
+    type
```
JGuetschow, line 17: Seems that my pycharm setting for docstrings is not correct, that added the colons, I think

primap2/csg/_strategies/gaps.py
```diff
@@ -71,15 +68,6 @@ class FitParameters:
         minimal number of points to calculate the trend. Default is 1, but if the degree
         of the fit polynomial is higher than 1, the minimal number of data points
         the degree of the fit polynomial
-
```
JGuetschow, line 74: So we use docstrings in the methods and not a methods section. I'll try to remember in the future.

JGuetschow (approved):

primap2/csg/_strategies/gaps.py
mikapfl: Yeah, both works, but generally I like the docs to be as close to the code as possible.

mikapfl:
@JGuetschow merge yourself if you are happy (will merge into your branch, not into main directly)

PR #313, opened 2025-02-18
 Ensure NaN values remain NaN during conversion

crdanielbusch:
# Pull request

Johannes found a bug in the convert function. All nan values were converted to zero. With these changes all nan are still nan after the conversion.

- [x] Tests added

## Description

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 -e .
-Sphinx>=4.2
+Sphinx>=4.2,<8.1.3
 myst-nb>=1
 sphinx-book-theme>=1.1
 numpydoc>=1.6
--- a/changelog/313.fix.md
+++ b/changelog/313.fix.md
@@ -1,0 +1,1 @@
+Fixed conversion of nan values.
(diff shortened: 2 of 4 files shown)


crdanielbusch:
> Nice, have you also checked that the new test fails with the old code?

Yes, the output will be all zeros.

docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 -e .
-Sphinx>=4.2
+Sphinx>=4.2,<8.1.3
```
mikapfl, line 2: nice catch!

crdanielbusch: There was a release that seems to break our docs. It's hard to figure out what the actual problem is. They are also dropping support for Python 3.10, but that shouldn't be the problem. Anything else we can do?

crdanielbusch: I tried to find some hints in the release notes: https://github.com/sphinx-doc/sphinx/releases

mikapfl: Looks like the autosummary accessors extension we use is not compatible with the latest release. I think pinning the version like you did is a good call. We can look at it again in 6 months or whenever we feel like it. We don't need any new features in the latest sphinx release.

PR #315, opened 2025-02-26
sparse array experiments

crdanielbusch:
# Pull request

- [ ] Tests added

## Description

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ wheels/
 .installed.cfg
 *.egg
 html/
+sparse_data/
 .mutmut-cache
 
 # PyInstaller
--- a/primap2/tests/conftest.py
+++ b/primap2/tests/conftest.py
@@ -54,6 +54,12 @@ def opulent_processing_ds() -> xr.Dataset:
     return examples._cached_opulent_processing_ds.copy(deep=True)
 
 
+@pytest.fixture
+def opulent_sparse_ds() -> xr.Dataset:
+    """Like the opulent dataset, but with sparse duck array."""
+    return examples._cached_opulent_sparse_ds.copy(deep=True)
+
+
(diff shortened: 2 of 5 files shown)


crdanielbusch:
@mikapfl @JGuetschow current state of our experiments on sparse arrays in primap2. All a bit messy, but I guess you get the idea. If we consider "merging with sparse arrays" is a minimal use case, I'd say sparse arrays are a promising approach. 

Notes:
- There is an issue with sparse array and xarray's `pr.set`, I've experienced the same issue as described here https://github.com/pydata/xarray/issues/9933, it works for pinned versions sparse 15.4.0 and xarray 2024.11.0
- This line in `_merge` causes errors: `da_error = da_comp.where(da_comp > tolerance, drop=False)` The xarray function `where` does not seem to be compatible with spars arrays.
- I think it is possible to move this test to the level of numpy arrays. What we would like to find out is essentially: Are there data points with higher errors than the specified tolerance.
- The tests in `test_merge.py` are passing, except for the ones that compare the error message. The error message with relevant the coordinates may be hard to put together, e.g. something like `pr.merge error: found discrepancies larger than tolerance (1.00%) `
- `xarray.Dataset.combine_first` (what we use for `pr.merge`) supports sparse arrays

mikapfl:
Quite promising. The advantage of the sparse array approach would of course be that conceptually, everything keeps working the same. The disadvantage is that the abstraction is leaky as we see with the `where` stuff. We'll have to keep "does this work with sparse arrays" always in mind. But maybe that is a fine trade-off, especially if we can get primap2 testing up to speed with sparse arrays and can guarantee that all primap2 functions work with sparse arrays.

Could you test if `pr.set` works on an array-level, e.g. setting a new country data?

PR #318, opened 2025-03-11
Support DataTree conversion

dnl-bsch:
Just an experiment to test the handling of DataTree objects.

This is what the convert method for DataTree objects would look like. Currently, convert is only implemented for DataArray objects. 

Background: Instead of merging sparse array we could consider to handle our data in xarray.DataTree objects for everything before the CSG in the primap-hist pipeline. To do this, we would have to adapt the primap2 functions to DataTree objects.

It seems to work quite well at the moment, but is based on an iteration through dataset and DataArray objects. Maybe we can utilise DataTree methods even more.

FYI @JGuetschow  @mikapfl


mikapfl:
Nice. If you wanted to avoid the manual for-loop, probably https://docs.xarray.dev/en/latest/generated/xarray.DataTree.map_over_datasets.html#xarray.DataTree.map_over_datasets is the right function to start with.

PR #322, opened 2025-03-19
Replace `xarray.core.ops` which is not part of the public xarray API

crdanielbusch:
# Pull request

Deals with https://github.com/primap-community/primap2/issues/320

- [x] Tests added

## Description

### DIFF
--- a/changelog/322.fix.md
+++ b/changelog/322.fix.md
@@ -1,1 +1,1 @@
-Replace xr.core.ops.fillna with fillna from public API
+Replace xr.core.ops.fillna with fillna from public PI
--- a/changelog/322.fix.md
+++ b/changelog/322.fix.md
@@ -1,0 +1,1 @@
+Replace xr.core.ops.fillna with fillna from public API
--- a/changelog/322.fix.md
+++ b/changelog/322.fix.md
@@ -1,1 +1,1 @@
-Replace xr.core.ops.fillna with fillna from public PI
+Replaced xr.core.ops.fillna with fillna from public xarray API for compatibility with upcoming xarray releases.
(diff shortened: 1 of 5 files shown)


crdanielbusch:
This turned out to be slightly more complicated. It seems like the option we set in `xr.set_options(arithmetic_join="exact")` does not apply to the `fillna` function. If we try to fill a time series with the years 1850 - 2022 with a time series 1851 - 2022, the value for 1850 would simply not be filled. We need to check before if the two array are exactly the same. There are different ways to do that. We could use `xr.align(ts, fill_ts, join="exact")` which would raise an error when indexes to be aligned are not equal. However, two time series with different names for the time axis would still go through

mikapfl:
> two time series with different names for the time axis would still go through

that seems fine

mikapfl:
Thanks for checking with an actual test and then finding a formulation which works as we want it to.

crdanielbusch:
Thanks for the input. I've added `xr.align` before all our `fillna`. Good intuition to double-check the set_options approach! ;)

mikapfl:
Please request a review when this is ready for merging from your side, thanks.

changelog/322.fix.md
```diff
@@ -0,0 +1 @@
+Replace xr.core.ops.fillna with fillna from public PI
```
mikapfl, line 1:
```suggestion
Replaced xr.core.ops.fillna with fillna from public xarray API for compatibility with upcoming xarray releases.
```

mikapfl (approved):
Thanks! Overall great speed for us that we now identify and solve issues with upcoming xarray releases within two weeks after the change was included in the xarray main branch. And good stuff with the tests.

PR #329, opened 2025-04-08
Add notes about data structures to documentation

crdanielbusch:
# Pull request

- [ ] Tests added

## Description

Add the notes on data structure research so they don't get lost

### DIFF
--- a/docs/source/data_structures.md
+++ b/docs/source/data_structures.md
@@ -32,9 +32,9 @@ We aim to retain as much of the current codebase as possible, though some parts 
 
 * We considered several approaches and chose [xarray datatree](https://docs.xarray.dev/en/stable/generated/xarray.DataTree.html).
 * Excluded approaches:
-  * [**Sparse array on top of numpy**](https://sparse.pydata.org/en/stable/): Memory-efficient but slow, with limited function optimization for our needs. Merge operations do not scale well.
+  * [**Sparse array on top of numpy**](https://sparse.pydata.org/en/stable/): Memory-efficient but slow, with limited function optimization for our needs. Merge operations do not scale well. Some experiments with sparse arrays [here](https://github.com/primap-community/primap2/pull/315/files).
   * **SQL database**: Could serve as a backend for data storage but is too slow for calculations and not suitable for our workflows.
-  * **Pandas**: Useful for pre-processing, where we already use it, but rewriting everything in pandas would be a significant effort. Mathematical operations are slow, and datasets remain sparse after converting back to xarray.
+  * **Pandas**: Useful for pre-processing, where we already use it, but rewriting everything in pandas would be a significant effort. Mathematical operations are slow, and datasets remain sparse after converting back to xarray. Basiv example of conversion in pandas [here](https://github.com/primap-community/primap2/pull/321/files).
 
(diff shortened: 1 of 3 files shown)


crdanielbusch:
@mikapfl I think these are the essential aspects of our research. Let me know if you think of anything else.

docs/source/data_structures.md
```diff
@@ -0,0 +1,138 @@
...
+    <td>
+      <ul>
+        <li>Limited examples and documentation available (as of April 2025).</li>
+        <li>Requires making our primap2 functions compatible with DataTree.</li>
+        <li>Processing DataTrees may involve looping through sub-datasets. Is it faster than merging once?</li>
+        <li>The hierarchical concept does not align well with our use case. All country datasets would be "siblings," but this is not necessarily a problem.</li>
+      </ul>
+    </td>
+  </tr>
+</table>
```
mikapfl, line 70:
```suggestion
| Pro | Con |
|---|---|
| Supports conversion from/to NetCDF and Zarr. | Limited examples and documentation available (as of April 2025). |
| Allows us to stay within the xarray ecosystem (as opposed to SQL or pandas). | Requires making our primap2 functions compatible with DataTree. |
| All pre-processing steps can be covered in DataTree.  | Processing DataTrees may involve looping through sub-datasets. Is it faster than merging once? |
| | The hierarchical concept does not align well with out use case. All country datasets would "siblings," but this is not necessarily a problem. |
```

It's not exactly the same result, but if possible avoid html and use markdown instead.

docs/source/data_structures.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Our specific requirements
+
+To handle very sparse arrays effectively, primap2 data handling functions should meet the following key requirements:
+
+* **Merge datasets**
+  Example: A script like `src/unfccc_ghg_data/unfccc_crf_reader/crf_raw_for_year.py` compiles all CRF (or CRT) datasets for a submission year into one dataset. Since countries report at different levels of detail, categories unique to one dataset result in NaNs in others, making datasets sparse and memory-intensive. We aim to significantly reduce memory usage.
+
+* **Select**
+  Filtering datasets by dimensions such as country, sector, or gas is essential. The `primap2` `.loc` function already supports this.
```
mikapfl, line 21:
```suggestion
  Filtering datasets by dimensions such as country, sector, or gas is essential. The {py:meth}`xarray.Dataset.pr.set` function already supports this, but not for DataTrees.
```
We can add a proper link using the `{py:meth}` syntax.

docs/source/data_structures.md
```diff
@@ -0,0 +1,138 @@
+### Data structures for primap2
+
```
mikapfl, line 2:
```suggestion

Here, we document an unfinished proposal for more advanced data structures for sparse data within primap2.
```

mikapfl (approved):
Nice, I have some small suggestions, but generally this should be useful if we later decide to pick it up again.

mikapfl:
Also, maybe "Data Structures for primap" is a bit too general as a title. Maybe "Sparse data ideas" instead, to make clear in the title already that this is not currently implemented functionality?

PR #330, opened 2025-04-14
Downscale based on given shares

crdanielbusch:
# Pull request

- [ ] Tests added

## Description

Our downscale function can only fill values, but doesn't have the option to add new basket contents to the dataset. For some use cases we need a downscaling function that calculates values based on shares from another dataset.

### DIFF
--- a/primap2/_downscale.py
+++ b/primap2/_downscale.py
@@ -164,13 +164,19 @@ class DataArrayDownscalingAccessor(BaseDataArrayAccessor):
         dim: Hashable,
         basket: Hashable,
         basket_contents: Sequence[Hashable],
-        basket_contents_shares: Sequence[float],
-        sel: dict[Hashable, Sequence] | None = None,
-    ) -> xr.Dataset:
-        da_sel = select_no_scalar_dimension(self._da, sel)
-        basket_da = da_sel.loc[{dim: basket}]
-        downscaled: xr.DataArray = basket_da * basket_contents_shares
-        return downscaled
+        basket_contents_shares: xr.DataArray,
+    ) -> xr.DataArray:
+        # TODO do we need this?
+        da = self._da.copy()
+
+        # normalise shares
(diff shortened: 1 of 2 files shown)


primap2/_downscale.py
```diff
@@ -158,6 +159,59 @@ def downscale_timeseries(
...
+        basket_contents_shares = basket_contents_shares / basket_contents_shares.sum(dim=dim)
+
+        downscaled = (
+            self._da.pr.loc[{dim: basket}] * basket_contents_shares.pr.loc[{dim: basket_contents}]
+        )
+
+        # xarray will try to match every indexed coordinate (coordinates with *)
+        # if they don't match the result will be empty, we need to make sure only
+        # the relevant coordinates are present
+        if downscaled.size == 0 or downscaled.isnull().all():
```
crdanielbusch, line 207: We need to make sure that there is overlap between `basket_contents_shares ` and `self._da`. It can easily happen that bot data arrays have the same indexed coordinates (e.g. `source`) with different values. As a result, `downscaled ` will be empty. It's not ideal to check if `downscaled` is empty, because there can be other reasons. Let me know if you have a better idea to prevent this case.

primap2/_downscale.py
```diff
@@ -158,6 +159,59 @@ def downscale_timeseries(
...
+        basket_contents_shares : xr.DataArray
+            The shares to use for downscaling.
+
+        Returns
+        -------
+        xr.DataArray
+            A new datarray with variables downscaled along `dim` using the provided shares.
+        """
+
+        # normalise shares
```
mikapfl, line 197:
```suggestion
        # normalise shares
        basket_contents_shares = basket_contents_shares.pr.loc[{dim: basket_contents}]
```
I think it makes sense to first select only the actual basket contents. If for example the reference data contain "1.A.1", "1.A.2" and "1.A.3", but basket_contents are only "1.A.1" and "1.A.2", then you would otherwise later not preserve the value of "1.A" fully.

PR #377, opened 2026-06-26
Data reading fixes

JGuetschow:
# Pull request

* A few changes in string processing. 
* Some failing tests fixed

Do not review or merge yet, I still have to test this thoroughly on actual data to see how much it messes up my data processing.

- [x] Tests added

## Description

* A few changes in string processing. Most importantly the order of mapping to 0 and NaN has changed for string containing multiple string codes. The reason ist that this is mostly used for higher level sectors where the list of codes is just the list of all codes in the subsectors. In this case it makes more sense to have NaN for the higher level sector if at least one of the subsectors is NaN instead of zero if at least one of the subsectors is zero.
* Some tests for the gaps functions in the csg were failing because the default time format has changed (us instead of ns). They have been fixed


fuchsi-huber:
I forgot to check for open PRs before starting with the fixes and upgrades. The test fixes are now already part of `main` - do you want the remaining changes from this PR merged as well @JGuetschow ?

PR #391, opened 2026-09-09
feat: added dataset convert_to_gwp functionality

### LINKED ISSUE #181: Add convert_to_gwp() to a dataset
**Is your feature request related to a problem? Please describe.**

We need to create a loop every time we want to convert the units for a data set.

**Describe the solution you'd like**

We would like to add an accessor for datasets for convert_to_gwp().

fuchsi-huber:
# Pull request

- [x] Tests added

## Description

Added the `convert_to_gwp` functionality for datasets requested in https://github.com/primap-community/primap2/issues/181 . Some decisions had to be made about how we handle inconsistent datasets:
* conversion replaces the original `da`s, contrary to the shape of some example datasets
* ~~round-trip conversion of one GWP to another is only performed if the additional argument `round_trip=True` is set - otherwise an exception is raised as this might indicate an accidental call or similar~~
* gas baskets cannot be converted so they are skipped. If they don't exist in the right GWP yet a warning is printed to notify the user of the inconsistency
* processing information is renamed to match the new variable names - but no new processing info is appended. Should this be changed to denote the conversion in the processing info instead? @mikapfl


docs/source/usage/select_and_view.md
```diff
@@ -174,6 +174,42 @@ ch4_sar = ds["CH4"].pr.convert_to_gwp_like(ds["CH4 (SARGWP100)"])
...
+
+list(converted.data_vars)
+```
+
+If the dataset contains gas baskets only in metrics you did not ask for, a warning is logged.
+
+For a *single gas* which is already given in a different global warming potential the
+conversion is possible by converting back to mass first. As this might indicate a logic error,
+it raises an error unless you specifically ask for it with `round_trip=True`.
+
```
mikapfl: I think "round trip" is the false mental image and confusing. You can convert between GHG metrics (GWPs) freely without having to convert to mass and back for single-gas species.

primap2/tests/test_units.py
```diff
@@ -80,6 +87,218 @@ def test_convert_to_mass_missing_info(opulent_ds: xr.Dataset):
...
+        # nothing is a global warming potential, so the dataset is unchanged
+        xarray.testing.assert_identical(converted, ds)
+        assert "Not converting" in caplog.text
+
+    def test_name_collision_raises(self, minimal_ds: xr.Dataset):
+        """SF6 (SARGWP100) converts to SF6, which already exists."""
+        with pytest.raises(
+            ValueError, match=r"Converting 'SF6 \(SARGWP100\)' would overwrite 'SF6'"
+        ):
+            minimal_ds.pr.convert_to_mass()
```
mikapfl, line 307: Seeing this, I think it would be super useful to have a mode or higher-level function which basically is "ensure that everything is available in this given GHG metric", i.e. which just skips stuff which already exists, instead of raising.

primap2/_units.py
```diff
@@ -302,3 +337,264 @@ def dequantify(self) -> xr.Dataset:
...
+            understood by ``openscm_units``.
+        units: str or pint unit
+            The units in which the global warming potential is given after the
+            conversion.
+        round_trip: bool, default False
+            How to treat single gases which are already given in a *different* global
+            warming potential. By default, an error is raised, because the conversion
+            can only be done by converting back to mass first. If ``True``, such
+            variables are converted back to mass and then to the requested global
+            warming potential.
```
mikapfl: Yeah, the API of openscm_units is not super useful here, so using it you technically have to convert to mass first then convert to the right GHG metric. But that is an implementation detail of openscm_units (maybe showing a fundamental API design problem in openscm_units, I'm not sure yet). Converting a single gas between different GHG metrics is trivial given the information we have in primap2 datasets (entities and gwp contexts). So this should be done always, not needing a special flag. And for how to implement it without the intermediate conversion (simply for performance), maybe we need a small helper which pulls conversion factors directly from https://github.com/openclimatedata/globalwarmingpotentials for a combination `entity, current_gwp_context, target_gwp_context`, where both `current_gwp_context` and `target_gwp_context` can also be `mass`?

mikapfl (changes requested):
Nice, very thorough conversion! Also, the global helper for processing information variable names is useful. 🙂 

I'm mainly confused by the `round_trip` parameter, and I think our users would as well.

docs/source/usage/select_and_view.md
fuchsi-huber: True. It did not fit my mental model of the intended usecase, but there really is no reason not to support it by default I guess :shrug:

primap2/tests/test_units.py
fuchsi-huber: For this I was really lacking a proper understanding of what the format was supposed to represent and how strict it should be. Is it a valid dataset if it has both a SF6 variable in mass units and one in gwp? The examples create these kinds of datasets but I don't see why any real-world dataset would be in such a state. We have no mechanism to ensure that they remain equivalent, right? Why would a user want this?

mikapfl: Hmm, I think such a dataset would be mostly useful simply for output, to make it easy for some random downstream consumer to pull it in their favourite GHG metric without having to compute it themselves.

I think a dataset with both is valid, but should only be generated if a user specifically requests it (by using da-level functions or special parameters). By default, we don't want duplication of information.

primap2/_units.py
fuchsi-huber: We are currently using pint to do both the unit and gwp conversion. We could do this as a two step process though to do the gwp conversion ourselves and leave the unit conversion to pint. I would prefer not to re-implement the unit conversion :sweat_smile:

primap2/tests/test_units.py
mikapfl: Maybe it raises to the level of an `info` or so in `ensure_valid` if some information is doubled (single gases in different GHG metrics)?

Or do you see a good reason to actually tighten the definition and enforce this in `ensure_valid`?

primap2/_units.py
mikapfl: Yes, we definitely wouldn't want to re-implement the unit conversion, but maybe we could stop relying on openscm_unit contexts given they are not expressive enough here.

mikapfl: Also, it warms my heart to only do unit conversion with the unit library, and do entity conversion somewhere else. It is of course a fundamental misunderstanding to put the entity in the unit (as in "Mt CO2"), and we can see that once you start it, it just scope-creeps. This conversion would work beautifully in openscm_units, if only we not only had the entity, but also the gwp_context in the unit (as in "Mt CO2_AR6GWP100"), or maybe also the original entity (as in "Mt SF6_as_CO2_AR6GWP100").

primap2/tests/test_units.py
fuchsi-huber: Okay, I would suggest we change the example datasets in that case, as they all contain something like ` minimal["SF6 (SARGWP100)"] = minimal["SF6"].pint.to("CO2 Gg / year")` in their construction. The documentation also highlights, that selecting by entity will select both `SF6` and `SF6 (SARGWP100)` in such a case and the test `test_add_aggregates_coordinates_result_sel` suggests that this is a thing that happens.... And I guess with different input sources this kind of makes sense - but maybe we should even enforce that this doesn't happen for a single source with a check in `ensure_valid`?

fuchsi-huber: > Or do you see a good reason to actually tighten the definition and enforce this in ensure_valid?

The reason to me would mostly be to ensure it is a well-defined dataset. If both are present I wouldn't know which one is more authoritative and which is derived or if they are even equivalent. (if they are: why should they both be present?)

mikapfl: Yeah, I agree. In that case, I don't think we want a carve-out for different sources. It is fine (and works) to have an inconsistent dataset, but before you `ensure_valid`, you have to fix that, and after you `ensure_valid` you can be sure that there is no trivially duplicated information. A well-formed dataset has all sources in the same GHG metric.

fuchsi-huber: At least for baskets we cannot quite ensure that. It makes sense even for a single dataset to contain the baskets in different metrics but I agree for all single gases. Should we check that the whole dataset is actually in the same metric or just that there are no redundancies?

mikapfl: I'd say, no redundancies. Gas baskets in different GHG metrics are actually additional information, in the general case.

primap2/_units.py
fuchsi-huber: I have implemented this via a detour using convert_to_mass for now. Using the gwp conversion factors directly would allow us to easily convert between GWP metrics, but if we also want to support converting from and to mass we would have to touch the units afterall which I don't want to do at the moment. The implementation as it is right now might be a little less efficient, but it is much better tested and thus more likely to be correct - and I don't think that this is a typical operation, so the performance isn't too important.

mikapfl (approved):

primap2/tests/test_units.py
fuchsi-huber: I have created the issue https://github.com/primap-community/primap2/issues/393 to track this change, so it doesn't block this PR.

PR #398, opened 2026-09-16
refactor: all our functions use the same skipna parameters

mikapfl:
# Pull request

- [x] Tests added

## Description

After some fixes, the `skipna` family of parameters (`skipna`, `skipna_evaluation_dims`, `min_count`) had different behaviour and defaults across our public functions. That's user-hostile (it is hard enough to understand what `skipna_evaluation_dims` does, no need to make it harder even by it varying across functions). So, now all our functions do the same thing.


mikapfl:
@fuchsi-huber If you review it, maybe focus on checking that the documentation now makes sense to you and matches what `sum` does. Everything else now follows `sum`, so if the documentation is understandable, this is a win.

